PR #16, opened 2020-08-08
Ghc 8 10

turion:
Merge #15 first


turion:
* [ ] Find out whether I need hvr at all

PR #26, opened 2020-08-16
Remove awkward tuple construction

turion:
Test performance (using speedtest & profiling)


turion:
Do Feedback as well

turion:
Testing the speedtest shows no appreciable difference between this and master. Slow down is < 5% and there are no spaceleaks introduced.

PR #48, opened 2020-11-25
Dev live program exceptions

turion:
* [ ] Clean up git
* [ ] Finish documentation
  * Refer to `CellExcept`
* [ ] Use `Void` to have variants of `safe` and `safely`


essence-of-live-coding/src/LiveCoding/LiveProgram/Except.hs
```diff
@@ -0,0 +1,54 @@
...
+  -> LiveProgram                  m
+foreverELiveProgram e LiveProgramExcept { .. } = liveCell $ foreverE e $ hoistCell commute $ runCellExcept unLiveProgramExcept
+
+foreverCLiveProgram
+  :: (Data e, Monad m)
+  => LiveProgramExcept m e
+  -> LiveProgram       m
+foreverCLiveProgram LiveProgramExcept { .. } = liveCell $ foreverC $ runCellExcept unLiveProgramExcept
+
+commute :: ExceptT e (ReaderT r m) a -> ReaderT r (ExceptT e m) a
```
turion, line 53: Should not be exported

essence-of-live-coding/src/LiveCoding/RuntimeIO/Launch.hs
```diff
@@ -32,8 +35,11 @@ class Monad m => Launchable m where
 instance Launchable IO where
   runIO = id
 
-instance Launchable (StateT (HandlingState IO) IO) where
-  runIO = runHandlingState
+instance (Typeable m, Launchable m) => Launchable (StateT (HandlingState m) m) where
+  runIO = runIO . runHandlingState
+
+instance (Data e, Finite e, Launchable m) => Launchable (ExceptT e m) where
+  runIO liveProgram = runIO $ foreverCLiveProgram $ tryLiveProgram liveProgram
```
turion, line 42: Maybe it makes more sense to stop the program upon exceptions. Looping forever is something the user can do themselves before they pass it to `runIO`.

turion: But this is not possible with the current definition of `Launchable`, since it translates to a `LiveProgram IO`, which can never terminate.

essence-of-live-coding/src/LiveCoding/RuntimeIO/Launch.hs
```diff
@@ -32,8 +35,11 @@ class Monad m => Launchable m where
 instance Launchable IO where
   runIO = id
 
-instance Launchable (StateT (HandlingState IO) IO) where
-  runIO = runHandlingState
+instance (Typeable m, Launchable m) => Launchable (StateT (HandlingState m) m) where
+  runIO = runIO . runHandlingState
+
+instance (Data e, Finite e, Launchable m) => Launchable (ExceptT e m) where
```
turion, line 41: Comment why the instance is as it is.

PR #51, opened 2021-04-07
Dev parametrised handle

turion:
Handles should be generalised so we can put parameters in the code that cause reinitialisation when changed. It's often possible to put such information on the type level, but quickly gets tedious.

* [x] Merge #48 first or rebase to get rid of some extraneous commits.
* [x] Clean up git history
* [x] Add tests
* [x] Do I want to make `Handle` an alias on top of `ParametrisedHandle`?


turion:
Merge #48 first.

PR #54, opened 2021-04-21
Add portmidi ~& vivid~ backends

turion:
## Do

* [x] Full documentation
* [ ] Some unit tests? Can I use/create MIDI devices on CI?
* [x] CI changes?
* [x] Clean up git history
* [x] `handlingPortMidiInput` doesn't yet reload on device change. Bug? Regression test? Do I need to use a parametrised handle?
* [x] `release` instead of `free`
* [x] Changelog
* [ ] Handles in control flow throw destructors and constructors.
  * Test this
  * This should be used instead of `SynthState`
* [x] Incorporate/adapt to https://github.com/miguel-negrao/eolc-vivid-midi-demo/pull/18

## Don't

* [x] Integration test? Integration tests are out of scope: #55

### DIFF
--- a/essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
+++ b/essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
@@ -102,3 +102,26 @@ portMidiInputWithProxy proxy = Handle
 
 handlingPortMidiInput :: (KnownSymbol name, MonadIO m) => Cell (HandlingStateT m) PortMidiHandle (Maybe (PortMidiInputStream (name :: Symbol)))
 handlingPortMidiInput = handling $ portMidiInputWithProxy Proxy
+
+data PortMidiDevices = PortMidiDevices
+  { inputDevices :: [DeviceInfo]
+  , outputDevices :: [DeviceInfo]
+  }
+
+getPortMidiDevices :: IO PortMidiDevices
+getPortMidiDevices = do
+  nDevices <- countDevices
+  devices <- mapM getDeviceInfo [0..nDevices-1]
+  return PortMidiDevices
+    { inputDevices = filter input devices
+    , outputDevices = filter output devices
+    }
+
+prettyPrintPortMidiDevices :: PortMidiDevices -> IO ()
+prettyPrintPortMidiDevices PortMidiDevices { .. } = do
+  putStrLn "\nPortMidi input devices:"
+  putStrLn $ unlines $ printName <$> inputDevices
+  putStrLn "\nPortMidi output devices:"
+  putStrLn $ unlines $ printName <$> outputDevices
+  where
+    printName dev = "- \"" ++ name dev ++ "\""
--- a/essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
+++ b/essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
@@ -1,3 +1,12 @@
+{- | * Support for [PortMidi](http://hackage.haskell.org/package/PortMidi)
+
+With this module, you can add cells which receive and send MIDI events.
+
+You don't need to initialise PortMidi, or open devices,
+this is all done by @essence-of-live-coding@ using the "LiveCoding.Handle" mechanism.
+-}
+
+{-# LANGUAGE Arrows #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
@@ -12,9 +21,12 @@
 module LiveCoding.PortMidi where
 
 -- base
+import Control.Concurrent (threadDelay)
 import Control.Monad (void, forM, join)
 import Control.Monad.IO.Class (MonadIO (liftIO))
 import Data.Either (fromRight)
+import Data.Foldable (traverse_, find)
+import Data.Function ((&))
 import Data.Maybe (catMaybes)
 import GHC.Generics
 import GHC.TypeLits (Symbol, symbolVal, KnownSymbol)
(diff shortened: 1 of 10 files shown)


turion:
@miguel-negrao What's a good latency to use for PortMidi's [`openOutput`](https://hackage.haskell.org/package/PortMidi-0.2.0.0/docs/Sound-PortMidi.html#v:openOutput)?

miguel-negrao:
> @miguel-negrao What's a good latency to use for PortMidi's [`openOutput`](https://hackage.haskell.org/package/PortMidi-0.2.0.0/docs/Sound-PortMidi.html#v:openOutput)?

SuperCollider uses 0, so I suggest doing the same:

https://github.com/supercollider/supercollider/blob/18c4aad363c49f29e866f884f5ac5bd35969d828/lang/LangPrimSource/SC_PortMIDI.cpp#L416

turion:
@miguel-negrao I found that having type-level annotations for midi devices is not much help in practice, and tends to make some things more complicated because we suddenly need `TypeApplications`. Is it fine with you if I drop them again, and replace them by parametrised handles? Then specifying a midi device is just a matter of an ordinary value-level string.

miguel-negrao:
The type-level anotations could in theory be propagated to the MIDI messages, and then it would allow functions such as 

```
cellCC :: Monad m => ControlCC -> Cell m (Maybe MidiMessage) (Maybe Float)
```

to be changed to something like

```
cellCC :: (KnownSymbol midiDeviceName, Monad m => ControlCC midiDeviceName -> Cell m (Maybe (MidiMessage midiDeviceName)) (Maybe Float)
```
Which wouldn't compile if you tried to filter midi messages from the wrong midi controller. But this would be tricky given that different OSes report different strings.

In practice this might complicate things and not bring much benefits. It's fine with me removing the type-level annotations.

turion:
Maybe the idea can be brought up again in https://github.com/turion/rhine, where it fits in naturally. eolc should be low-level, and thus not have so many complicated type level constructs.

miguel-negrao:
I had a look at latest changes. It's looking good. I have a question:

In `readEventsFrom` the output type is `[PMEvent]`. It will be very cumbersome to have to map over that list in all midi code. The ideal would be to get rid of the list and fire each event separately. Would it be possible within the constraints of eolc to create a function `Cell m1 a [b] -> Cell m2 a b` ? One approach I can imagine is to create a queue in the monad, for instance using StateT. Here is an attempt which I tested and seems to work. This does add another layer to the monad transformer stack.

```Haskell
main :: IO ()
main = foreground liveProgram

liveProgram :: LiveProgram IO
liveProgram = liveCell $ cellTest (runStateC_ (flattenList cell1) [])

cell1 :: Monad m => Cell m [Int] [Int] 
cell1 = arr id

cellTest :: (Read a, Show b) => Cell IO a b -> Cell IO () ()
cellTest cell =
    arrM (const (read <$> getLine)) >>>
    cell >>>
    arrM print

flattenList ::  Monad m => Cell m a [b] -> Cell (StateT [b] m) a (Maybe b) 
flattenList cell = liftCell cell >>> arrM f where
  f list = do
    queue <- get
    case queue of
      [] -> do
        case list of
          [] -> return Nothing
          (x:xs) -> do
            put xs
            return $ Just x
      (y:ys) -> do
        put (ys ++ list)
        return $ Just y
```
testing:
```Haskell
[1,2]
Just 1
[]
Just 2
[]
Nothing
```

turion:
Yes, something like this is possible. You don't even need to use `StateT` in the monad, you can use the internal state of the cell. `dunai` has this: http://hackage.haskell.org/package/dunai-0.7.0/docs/src/Data.MonadicStreamFunction.Util.html#fifo And I think you can copy the code without change and use it as a cell.

turion:
(Although an efficient implementation will use a `Sequence` instead of a list)

turion:
I added a commit containing such a `fifo` buffer. Does it simplify the situation? You might need https://hackage.haskell.org/package/containers-0.6.3.1/docs/Data-Sequence.html#v:fromList

miguel-negrao:
> I added a commit containing such a `fifo` buffer. Does it simplify the situation? You might need https://hackage.haskell.org/package/containers-0.6.3.1/docs/Data-Sequence.html#v:fromList

Sure, I think that would work. I couldn't test because currently eolc is not compiling here (see [1], travis reports the same error), but maybe this would work ? 

```Haskell
readEventsFrom
  :: MonadIO m
  => Cell (PortMidiT m) PortMidiInputStream (Maybe PMEvent)
readEventsFrom = (arrM $ (fromList <$>) . liftPMError . liftIO . readEvents . unPortMidiInputStream
  >>> fifo
```

[1]
```
stack build
essence-of-live-coding                > configure (lib + exe)
essence-of-live-coding                > Configuring essence-of-live-coding-0.2.5...
essence-of-live-coding                > build (lib + exe)
essence-of-live-coding                > Preprocessing library for essence-of-live-coding-0.2.5..
essence-of-live-coding                > Building library for essence-of-live-coding-0.2.5..
essence-of-live-coding                > [34 of 40] Compiling LiveCoding.Handle
essence-of-live-coding                > 
essence-of-live-coding                > /home/miguel/Development/Haskell/projects/SC/essence-of-live-coding/essence-of-live-coding/src/LiveCoding/Handle.hs:146:11: error:
essence-of-live-coding                >     • Couldn't match expected type ‘h
essence-of-live-coding                >                                     -> Control.Monad.Operational.ProgramT
essence-of-live-coding                >                                          (HandlingStateInstruction m) m a0’
essence-of-live-coding                >                   with actual type ‘Control.Monad.Operational.ProgramT
essence-of-live-coding                >                                       (HandlingStateInstruction m) m ()’
essence-of-live-coding                >     • The function ‘reregister’ is applied to three arguments,
essence-of-live-coding                >       but its type ‘m ()
essence-of-live-coding                >                     -> Data.IntSet.Internal.Key
essence-of-live-coding                >                     -> Control.Monad.Operational.ProgramT
essence-of-live-coding                >                          (HandlingStateInstruction m) m ()’
essence-of-live-coding                >       has only two
essence-of-live-coding                >       In a stmt of a 'do' block:
essence-of-live-coding                >         reregister
essence-of-live-coding                >           (destroyParametrised parameter mereHandle) key mereHandle
essence-of-live-coding                >       In the expression:
essence-of-live-coding                >         do mereHandle <- lift
essence-of-live-coding                >                            $ changeParametrised lastParameter parameter mereHandle
essence-of-live-coding                >            reregister
essence-of-live-coding                >              (destroyParametrised parameter mereHandle) key mereHandle
essence-of-live-coding                >            return
essence-of-live-coding                >              (mereHandle, Handling {handle = (mereHandle, parameter), ..})
essence-of-live-coding                >     • Relevant bindings include
essence-of-live-coding                >         mereHandle :: h (bound at src/LiveCoding/Handle.hs:145:11)
essence-of-live-coding                >         handling :: Handling (h, p)
essence-of-live-coding                >           (bound at src/LiveCoding/Handle.hs:140:14)
essence-of-live-coding                >         cellStep :: Handling (h, p)
essence-of-live-coding                >                     -> p
essence-of-live-coding                >                     -> Control.Monad.Operational.ProgramT
essence-of-live-coding                >                          (HandlingStateInstruction m) m (h, Handling (h, p))
essence-of-live-coding                >           (bound at src/LiveCoding/Handle.hs:135:5)
essence-of-live-coding                >         destroyParametrised :: p -> h -> m ()
essence-of-live-coding                >           (bound at src/LiveCoding/Handle.hs:132:33)
essence-of-live-coding                >         changeParametrised :: p -> p -> h -> m h
essence-of-live-coding                >           (bound at src/LiveCoding/Handle.hs:132:33)
essence-of-live-coding                >         createParametrised :: p -> m h
essence-of-live-coding                >           (bound at src/LiveCoding/Handle.hs:132:33)
essence-of-live-coding                >         (Some bindings suppressed; use -fmax-relevant-binds=N or -fno-max-relevant-binds)
essence-of-live-coding                >     |
essence-of-live-coding                > 146 |           reregister (destroyParametrised parameter mereHandle) key mereHandle
essence-of-live-coding                >     |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
essence-of-live-coding                > 
Progress 1/12

--  While building package essence-of-live-coding-0.2.5 (scroll up to its section to see the error) using:
      /home/miguel/.stack/setup-exe-cache/x86_64-linux-tinfo6/Cabal-simple_mPHDZzAJ_3.0.1.0_ghc-8.8.4 --builddir=.stack-work/dist/x86_64-linux-tinfo6/Cabal-3.0.1.0 build lib:essence-of-live-coding exe:TestExceptions exe:TestNonBlocking --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
```

turion:
>  I couldn't test because currently eolc is not compiling here

Whoops, I forgot to push the fix, sorry. Pushed now.

miguel-negrao:
Still not quite compiling: 
```
essence-of-live-coding-speedtest-yampa> /home/miguel/Development/Haskell/projects/SC/essence-of-live-coding/essence-of-live-coding-speedtest-yampa/app/SpeedTest.hs:7:1: error: parse error on input ‘import’
essence-of-live-coding-speedtest-yampa>   |
essence-of-live-coding-speedtest-yampa> 7 | import Control.Arrow
essence-of-live-coding-speedtest-yampa>   | ^^^^^^
essence-of-live-coding-speedtest-yampa> 
```

Is `module <X> where` not needed  ?

turion:
Weird, I don't have that build error. I cleaned up the git history a bit. Can you check again?

miguel-negrao:
> Weird, I don't have that build error. I cleaned up the git history a bit. Can you check again?

Still not compiling. It can be something on my end, I will investigate. I'm compiling with stack, where I have a stack file like this in the root of the folder.  I see you have some stack files with resolver lts-13.19, are you using those or are you using just nix and cabal ? Regarding nix, which nixpks commit are you using ?

```
resolver: lts-16.23

packages:
- demos
- essence-of-live-coding
- essence-of-live-coding-gloss
- essence-of-live-coding-gloss-example
- essence-of-live-coding-pulse
- essence-of-live-coding-pulse-example
- essence-of-live-coding-quickcheck
- gears
- essence-of-live-coding-PortMidi
- essence-of-live-coding-speedtest-yampa
- essence-of-live-coding-vivid
- essence-of-live-coding-warp

extra-deps:
- PortMidi-0.2.0.0@sha256:0671e36ec72e95138bf396234b205864a8a6d0ee353e09e01cbfd57004c56f40,2383
- midi-simple-0.1.0.0@sha256:06d400db456ec07bf8160858024f866eee346d90effc32ec1a2d819878df37f0,2405
- vivid-0.5.1.0@sha256:fb8b2055ed3d13e267bd2d92052e659ff1a6bc11f2d9d4a753094d35204353ec,8330
- vivid-osc-0.5.0.0@sha256:633e5c1c87633feaa26019442c3ae63946001c7afe2afa4d2cf22586bd4b182c,2786
- vivid-supercollider-0.4.1.2@sha256:c7493f2a67b04589df8943e58a7445803e3365c6ad439603bf35695734b9bd65,1738
- Yampa-0.13.1@sha256:4612a2646c27bcd3ac55c90dbc34249303e28aa5b3bc3e0c6fa9ce58b889843c,5436
```

turion:
Only cabal is supported currently. The stack files are leftovers or accidental commits.

turion:
> Is `module <X> where` not needed ?

No, it's the main executable.

miguel-negrao:
Will try to have a look at the issue of why it is not compiling later, but for now I don't need that package.

I'm looking at the new code, but I haven't been able to figure out how to combine `loopPortMidiC` and `readEventsC` .  `loopPortMidiC` needs a `PortMidiT m` but `readEventsC` is `(HandlingStateT (PortMidiT m))`. How should these two cells be combined ?

turion:
> Will try to have a look at the issue of why it is not compiling later, but for now I don't need that package.

Yes, it's just an old test.

> I'm looking at the new code, but I haven't been able to figure out how to combine `loopPortMidiC` and `readEventsC` . `loopPortMidiC` needs a `PortMidiT m` but `readEventsC` is `(HandlingStateT (PortMidiT m))`. How should these two cells be combined ?

That step is indeed what I've been working on these days. It's surprisingly complicated. One downside of using transformers so much is that an early poor choice of stacking order bites you later, and that's what happened here. I have an idea to get out of it, will try that later.

turion:
@miguel-negrao Ok, I've reworked that part. Now it should be possible to use and run the portmidi cells. The easiest approach will be with `writeEventsC` and `readEventsC`.

miguel-negrao:
@turion I've reworked the code in eolc-midi-demo to use your latest commit. You can find that code [here](https://github.com/miguel-negrao/eolc-vivid-midi-demo/tree/dev_vivid_handle_another_attempt/eolc-midi-demo) (branch dev_vivid_handle_another_attempt).  I've refactored code a bit, putting all common code in Lib.hs.

I'm getting the following error when running `sc-note-demo`: 

```
sc-note-demo: src/LiveCoding/Exceptions.lhs:(108,1)-(119,36): Non-exhaustive patterns in function runExceptC
```

turion:
> I'm getting the following error when running `sc-note-demo`:
> 
> ```
> sc-note-demo: src/LiveCoding/Exceptions.lhs:(108,1)-(119,36): Non-exhaustive patterns in function runExceptC
> ```

Whoops! Thanks for finding. That's a bug in my library. Reported here: https://github.com/turion/essence-of-live-coding/issues/63

turion:
> Also, I've tried using `fifo` but it needs a `Data` instance for `PMEvent` and that requires a `Data` instance for `Timestamp` which is not exported by PortMidi apparently. The monad version I proposed doesn't have this issue I suppose, but it will lose events on reload, right ?

It depends on which version you mean exactly. As a rule of thumb, all livecodeable state needs a `Data` constraint. [`runStateC_`](https://hackage.haskell.org/package/essence-of-live-coding-0.2.5/docs/LiveCoding.html#v:runStateC_) needs the `Data` constraint as well, since it puts the state from the monad into a cell.

Your `flattenList` doesn't need the constraint at first glance:
```haskell
flattenList ::  Monad m => Cell m a [b] -> Cell (StateT [b] m) a (Maybe b) 
```
But now you need to wonder how to run it. If you use `runStateC_`, you'll need `Data b` again. You could do something ugly and use [`ioRefHandle`](https://hackage.haskell.org/package/essence-of-live-coding-0.2.4/docs/LiveCoding-Handle-Examples.html#v:ioRefHandle) to store the state. But I think it's nicer if you just add the orphan `Data` instance and maybe send a PR to upstream to add the missing instance.

turion:
I fixed the bug you encountered.

miguel-negrao:
The issue is gone. Thanks !

Now I get

```
Encountered PortMidi exception:
NotAnInputDevice
```
I think `lookupDeviceID` and `portMidiInputStreamHandle` have an issue. In `lookupDeviceID` you are finding the first element that matches the name (`find ((nameLookingFor ==) . name . fst) devices`) but that element might not be of the right type, for instance it might be an output device, when we want the input device with the same name. In the case of my system the first device with name I'm looking for is of output type, the one with input type comes after.

turion:
I erroneously assumed that the name was unique. Fixed now, I hope. You're a very good tester.

miguel-negrao:
Ok, now everything is working again. Nice !

Regarding `readEventsC`, it outputs a PMEvent which is very low level. Do you plan on supporting output as higher level MidiMessage from the simple-midi package ?

turion:
> Regarding `readEventsC`, it outputs a PMEvent which is very low level. Do you plan on supporting output as higher level MidiMessage from the simple-midi package ?

No, that should be a separate package. I believe you already have some code in that direction. Maybe we can make that into a library `PortMidi-simple`, and then use it from here?

miguel-negrao:
> No, that should be a separate package. I believe you already have some code in that direction. Maybe we can make that into a library `PortMidi-simple`, and then use it from here?

It's just a single function, 4 lines of code ( `pmEventToSP`). That seems rather small for a package. Perhaps I can add it to the `sound-controllers-eolc` package, and that would be the high level approach to eolc midi, since it also provides abstractions for physical controllers.

turion:
> It's just a single function, 4 lines of code ( `pmEventToSP`). That seems rather small for a package.

I see no problem with that in principle, but I can understand it's too much trouble. The only thing I'm sure about is that I don't want to add more than one backend dependency to every `essence-of-live-coding-*` package.

> Perhaps I can add it to the `sound-controllers-eolc` package, and that would be the high level approach to eolc midi, since it also provides abstractions for physical controllers.

That sounds like a good idea.

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,200 @@
...
+portMidiInputStreamHandle
+  :: MonadIO m
+  => String
+  -> Handle m (Either EOLCPortMidiError PortMidiInputStream)
+portMidiInputStreamHandle name = Handle
+  { create = runExceptT $ do
+      deviceID <- ExceptT $ lookupDeviceID name Input
+      fmap PortMidiInputStream $ withExceptT PMError $ ExceptT $ liftIO $ openInput deviceID
+  -- TODO I don't get the error from closing here.
+  -- Actually I really want ExceptT in the monad
```
turion, line 130: Follow-up issue

miguel-negrao:
@turion perharps you can rebase this on master so that I can make a pull-request for stack support also on this branch, if that is okay ?

turion:
> @turion perharps you can rebase this on master so that I can make a pull-request for stack support also on this branch, if that is okay ?

I tried to add stack support here, can you have a look whether everything is fine for you?

miguel-negrao:
> I tried to add stack support here, can you have a look whether everything is fine for you?

Yes, it is building correctly here. Thanks !

turion:
I'm changing the scope of this MR to only PortMidi so I can merge soon. I'll continue with vivid separately.

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,262 @@
...
+--   Typically needed to lift PortMidi backend functions.
+liftPMError :: Monad m => m (Either PMError a) -> PortMidiT m a
+liftPMError = PortMidiT . ExceptT . fmap (left PMError) . lift
+
+-- | Given a cell with existing handles, lift it into 'PortMidiT'.
+liftHandlingState :: Monad m => Cell (HandlingStateT m) a b -> Cell (PortMidiT m) a b
+liftHandlingState = hoistCell $ PortMidiT . lift
+
+-- Trouble is I really want to go from PortMidiT (ExceptT e m)
+joinCellExcept :: (Data e, Finite e, Monad m) => CellExcept (PortMidiT m) a b e -> CellExcept (HandlingStateT m) a b (Either e EOLCPortMidiError)
```
turion, line 109: Unfinished. Put in separate PR

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,262 @@
...
+2. Run @cell@, until possibly an exception occurs
+3. Shut the MIDI system down
+4. Throw the exception in 'CellExcept'
+-}
+runPortMidiC :: MonadIO m => Cell (PortMidiT m) a b -> CellExcept (HandlingStateT m) a b EOLCPortMidiError
+runPortMidiC cell = try $ proc a -> do
+  _ <- liftCell $ handling portMidiHandle -< ()
+  hoistCell unPortMidiT cell -< a
+
+loopPortMidiC :: MonadIO m => Cell (PortMidiT m) a b -> Cell (HandlingStateT m) a b
```
turion, line 127: Document

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,262 @@
...
+You will rarely need this function.
+Look at 'runPortMidiC' and 'loopPortMidiC' instead.
+-}
+runPortMidiT :: PortMidiT m a -> HandlingStateT m (Either EOLCPortMidiError a)
+runPortMidiT PortMidiT { .. } = runExceptT unPortMidiT
+
+-- | A marker witnessing that PortMidi was initialized
+data PortMidiHandle = PortMidiHandle
+
+portMidiHandle :: MonadIO m => Handle m PortMidiHandle
```
turion, line 152: Still needed? Document? Hide?

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,262 @@
...
+portMidiHandle = Handle
+  { create = do
+      liftIO initialize
+      return PortMidiHandle
+  , destroy = const $ liftIO $ void terminate
+  }
+
+newtype PortMidiInputStream = PortMidiInputStream { unPortMidiInputStream :: PMStream }
```
turion, line 160: Document

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,262 @@
...
+newtype PortMidiOutputStream = PortMidiOutputStream { unPortMidiOutputStream :: PMStream }
+
+data InputOrOutput = Input | Output
```
turion, line 163: Document

turion (commented):
Lots of docs missing

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
```diff
@@ -0,0 +1,287 @@
...
+{- | Exceptions that can occur while doing livecoding with PortMidi.
+
+There are two kinds of exceptions:
+
+* Internal PortMidi exceptions (see 'PMError')
+* When a device is not correctly specified by name and input/output configuration
+-}
+data EOLCPortMidiError
+  -- | An internal error occurred in the PortMidi library
+  = PMError PMError
```
turion, line 71: Not sure that haddocks correctly

turion (commented):
Haddock

essence-of-live-coding-PortMidi/src/LiveCoding/PortMidi.hs
turion: I'm not even sure right now what exactly the issue is. I'll let it rest and finish the Handle refactorings first.

PR #61, opened 2021-05-24
Dev handling state refactor

turion:
* [ ] Check whether this allows progress in #54 
* [ ] Document fully
* [ ] Spend a thought on what is public interface and what not (probably the HandlingState module should be private (`other-modules`?), but testable


turion:
`operational` should not be needed. Instead, `HandlingStateT` can be more efficiently encoded as `AccumT` and `WriterT`.

turion:
It turns out this is not needed for #54.

PR #74, opened 2021-06-25
liveSynth - avoid having to explicitly create synthdef

miguel-negrao:
The defaults passed to the sd function are only used when first sending the synthdef. For creating the synth the params var is used. The default values for the sd function can therefore be derived from params by taking their value on the first tick. This ensures they will not change further which will avoid sending the synthdef again. Only changing sdbody will cause the synthdef to be sent again to scsynth.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
       - libpulse-dev
       - libblas-dev
       - liblapack-dev
+      - libasound2-dev
     sources:
       - sourceline: 'ppa:hvr/ghc'
 cache:
--- a/README.md
+++ b/README.md
@@ -211,3 +211,8 @@ it's advisable to follow these patterns:
   (at least version 3).
   If this is not the case for you,
   consider updating, or adapt the files to use `new-`style commands.
+
+## Contributors
+
+* Manuel Bärenz https://github.com/turion/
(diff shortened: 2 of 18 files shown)


turion:
Can you reopen, rebase onto `master` and point it onto the branch `dev_vivid`?

miguel-negrao:
> Can you reopen, rebase onto `master` and point it onto the branch `dev_vivid`?

Sure, will do that.

miguel-negrao:
I don't think I can reopen the pull-request, perhaps I don't have privileges ? I cherry picked the commit on top of dev-vivid, that seemed easier to do. By the way dev-vivid is not building currently, so I couldn't test the PR.

turion:
That's probably because the target branch doesn't exist anymore. Maybe it's best when you target it towards `master` right away. Then I'll merge #77 first (and possibly cherry-pick whatever is necessary).

miguel-negrao:
I seem to not be able to change the target either, when I click the "edit" button I don't see the target pull-down menu:

imagem

Should I open a new PR ?

turion:
Weird. Ok, then a new PR is the right course of action.

miguel-negrao:
New PR done.

PR #75, opened 2021-06-29
Add nix support for stack

turion:
CC @miguel-negrao


turion:
Otherwise it will not build on a NixOS machine.

PR #83, opened 2021-07-09
liveSynth - avoid having to explicitly create synthdef v2

miguel-negrao:
The defaults passed to the sd function are only used when first sending the synthdef. For creating the synth the params var is used. The default values for the sd function can therefore be derived from params by taking their value on the first tick. This ensures they will not change further which will avoid sending the synthdef again. Only changing sdbody will cause the synthdef to be sent again to scsynth.

### DIFF
--- a/essence-of-live-coding-vivid/src/LiveCoding/Vivid.hs
+++ b/essence-of-live-coding-vivid/src/LiveCoding/Vivid.hs
@@ -6,7 +6,9 @@ With this module, you can create cells corresponding to synthesizers.
 The synthesizers automatically start and stop on reload.
 -}
 
+{-# LANGUAGE Arrows #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveDataTypeable #-}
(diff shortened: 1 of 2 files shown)


essence-of-live-coding-vivid/src/LiveCoding/Vivid.hs
```diff
@@ -39,8 +42,18 @@ vividHandleParametrised = ParametrisedHandle { .. }
     changeParametrised old new synth = defaultChange createParametrised destroyParametrised old new synth
 
 deriving instance Eq (SynthDef args)
+deriving instance Data SynthState
+deriving instance KnownSymbol a => Data (I a)
 
 liveSynth
```
turion, line 48: @miguel-negrao Do you have a good usage example for this function?

miguel-negrao: All examples in eolc-vivid-midi-demo (in eolc-midi-demo/app/) use this function except for 
SC_Embedded.hs.

turion:
Thanks!

PR #89, opened 2021-12-30
Add foreverCE

turion:


miguel-negrao:
I find the following function useful also. `foreverE` requires use of the `Reader` monad, which requires lifting cells with a specific monad into the Reader layer. Often there is no need to send the exception value deep into the cell hierarchy via Reader, and using the exception value as extra input is cleaner (less use of `liftCell`).

```Haskell
foreverE' ::
    (Monad m, Data e) =>
    e -> Cell (ExceptT e m) (e,a) b -> Cell m a b
foreverE' e cell = foreverE e (readerToInput cell) where
    readerToInput cell = proc a -> do
        e <- constM ask -< ()
        liftCell cell -< (e,a)
```

My big breakthrough in understanding foreverE was when I understood that, actually, there is no Finite requirement on the exception type as long as we don't try to use exception to determine the "shape" of the cell, and just use it as an input to the cell. I suspect many recursive switching definitions are of this type. Together with `resampleListPar` which dynamically creates n cells based on input list size, this already allows to cover situations where the the switching changes the number of cells but not the general "shape" of the cells. Still, a continuation based approach without the Finite requirement would obviously be more  powerful.

essence-of-live-coding/src/LiveCoding/Forever.lhs
```diff
@@ -141,3 +142,20 @@ Let us run it:
...
+\begin{comment}
+\begin{code}
+foreverCE :: (Monad m, Data e, Finite e) =>
+  e ->
+  (e -> CellExcept a b m e) ->
+  Cell m a b
+foreverCE e f = foreverE e $ commute $ runCellExcept . f
+
+foreverCE' ::
+  (Monad m, Data void, Finite void, Data e, Finite e) =>
```
turion, line 155: I could avert the constraints on `void` by plugging in `Void` instead. Weird it's necessary at all. Have a look again whether I can get rid of them, and write a test just to see whether they are a nuisance.

turion:
Yes, I should add that function.

PR #87, opened 2021-09-13
Set up Github Workflows

turion:

PR #91, opened 2022-01-11
Add github actions for CI

miguel-negrao:
- update vivid to 0.5.2.0 in order to build with latest hashable. This version adds an Eq instance for SynthDef, which therefore doesn't need to be defined in this repo.

Solves #87


miguel-negrao:
Strange, the pull-request didn't trigger the action in this repo... Perhaps it's a security issue ?

miguel-negrao:
> Strange, the pull-request didn't trigger the action in this repo... Perhaps it's a security issue ?

Ok, needed to add "pull_request".

miguel-negrao:
Doesn't build with ghc 9.2, some constraints would have to be relaxed I think. See here: https://github.com/miguel-negrao/essence-of-live-coding/runs/4780786282?check_suite_focus=true

miguel-negrao:
Ok, it builds with cabal with 8.8 . I'll add that.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+          ghc-version: ${{ matrix.ghc }} # Resolves to the latest point release of GHC 8.8
+          cabal-version: '3.6.2.0' # Exact version of Cabal
+          enable-stack: true
+          stack-version: '2.7'
+      - name: Install dependencies
+        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev
+      - name: Cabal update
+        run: cabal update
+      - name: Freeze Cabal plan
+        run: cabal freeze
```
turion, line 25: I wonder whether `cabal freeze` should instead be done manually, and the result committed to the repo.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Cache Cabal build artifacts
+        uses: actions/cache@v2
+        with:
+          path: |
+            ${{ steps.setup-haskell-build.outputs.cabal-store }}
+            dist-newstyle
+          key: ${{ runner.os }}-cabal-${{ matrix.ghc }}-${{ hashFiles('cabal.project.freeze') }}
```
turion, line 32: Ok, that is pretty cool, hashing on the result of the cabal freeze :) :+1:

essence-of-live-coding-vivid/src/LiveCoding/Vivid.hs
```diff
@@ -58,7 +58,6 @@ vividHandleParametrised = ParametrisedHandle { .. }
     -- Synthdef or start/stop state changed, need to release and reinitialise
     changeParametrised old new synth = defaultChange createParametrised destroyParametrised old new synth
 
-deriving instance Eq (SynthDef args)
```
turion, line 61: Why this change?

turion: Nevermind, I didn't read your message completely it seems.

turion:
Does adding GHC 9.0 work?

essence-of-live-coding-vivid/src/LiveCoding/Vivid.hs
miguel-negrao: vivid was not building anymore with the version of hashable picked by cabal because it needed SynthDef to be an instance of `Eq`. I made a commit to fix that, and sent a pull-request to Tom, which he accepted: https://git.sr.ht/~amindfv/vivid/commit/2b340aa5c57dc5a0d57e55943b38fc92c9c882e3 . He then bumped vivid to 0.5.2.0.

Since the Eq instance is now in Vivid it is not needed here.

.github/workflows/ci.yml
miguel-negrao: I don't have much experience lately with cabal, but I think that would make the builds more deterministic, making it more similar to stack in that respect, so I think it might be a good idea.

.github/workflows/ci.yml
miguel-negrao: Can't claim credit, it's all copy-paste from various blog posts. :-)

miguel-negrao:
> Does adding GHC 9.0 work?

No. I think the constraints on some packages probably need to be changed. Check the logs here: https://github.com/miguel-negrao/essence-of-live-coding/runs/4780786282?check_suite_focus=true

miguel-negrao:
Ok, I have 8.8.4 and 8.10.7 working: https://github.com/miguel-negrao/essence-of-live-coding/runs/4781303652?check_suite_focus=true

miguel-negrao:
> > Does adding GHC 9.0 work?
> 
> No. I think the constraints on some packages probably need to be changed. Check the logs here: https://github.com/miguel-negrao/essence-of-live-coding/runs/4780786282?check_suite_focus=true

9.0.2 also fails, with problems in eolc code: https://github.com/turion/essence-of-live-coding/runs/4781676336?check_suite_focus=true

I suggest for now start with 8.8 and 8.10.

miguel-negrao:
Added a cabal flag to show the result of the tests

.github/workflows/ci.yml
turion: Can you rebase on `master` and see whether it works without this step?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          cabal build all --enable-tests
+      - name: Cabal test
+        run: |
+          cabal test all --enable-tests --test-show-details=Always
+      - name: Cache stack build artifacts
+        uses:  freckle/stack-cache-action@v1.0.1 
+        with:
+          stack-yaml: stack.yaml
+      - name: Stack build and test (lts-18.21 - ghc 8.10.7)
```
turion, line 48:
```suggestion
      - name: Stack build and test (lts-18.21)
```

(GHC version is part of the matrix)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Cabal build dependencies
+        run: |
+          cabal build all --enable-tests --only-dependencies
+      - name: Cabal build all
+        run: |
+          cabal build all --enable-tests
+      - name: Cabal test
+        run: |
+          cabal test all --enable-tests --test-show-details=Always
+      - name: Cache stack build artifacts
```
turion, line 44: There is no clever way to have a cache key depend on the stack lock file? Would it be a good idea to have that?

miguel-negrao: I believe it already does. Looking at the source of the action [here](https://github.com/freckle/stack-cache-action/blob/main/src/stack-cache.js) it will hash the lock file, the cabal files and the source code. It uses the latest cache available from the more specific to the least specific. So if the source changes but the cabal files remain the same, it uses latest cache available with same cabal files, if cabal changes it uses the latest cache with the same lock file and if the lock file changes it does everything from scratch.

.github/workflows/ci.yml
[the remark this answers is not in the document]
miguel-negrao: I don't have a lot insight into that, but I don't think it is needed. If we did that then we should also have a matrix over cabal versions. For instance on [this repo](https://github.com/kowainik/summoner/blob/main/.github/workflows/ci.yml) they use a single cabal and stack version each. Our stack files don't use anything advanced only available on the latest version, it should work fine on older versions I believe, but if we want to be sure we can always run it once with a matrix over stack and see what is the result.

miguel-negrao:
There's an issue with the version of base in the `cabal.project.freeze` ...

miguel-negrao:
I think the issue is that we need a different cabal.project and cabal.project.freeze file for each version of ghc, similarly to stack. That can be done as explained [here](https://github.com/haskell/cabal/issues/736). Will try to look at it tomorrow. Curiously the freeze files can be created [from stackage](https://medium.com/@fommil/why-not-both-8adadb71a5ed) for the required ghc version.

.github/workflows/ci.yml
turion: Right! That's great.

.github/workflows/ci.yml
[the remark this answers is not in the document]
turion: Sorry, I actually meant different _stackage_ versions (i.e. not only lts-18.21, but also the older ones).

turion:
> I think the issue is that we need a different cabal.project and cabal.project.freeze file for each version of ghc, similarly to stack. 

Maybe I was wrong about the freeze files then. Maybe they simply shouldn't be committed.

I asked that on reddit now: https://www.reddit.com/r/haskell/comments/s4fcoy/should_i_commit_cabal_freeze_files_to_my_repo/?

For now, is it possible to simply overwrite the freeze file by calling `cabal freeze` again on CI? (Pretty much reverting to the state it was before I said to remove the `cabal freeze` step?)

> That can be done as explained [here](https://github.com/haskell/cabal/issues/736).

Hmm, that issue doesn't seem to talk about `cabal freeze`. Sure it's the one you wanted to link?

miguel-negrao:
The freeze file will also pin the base version to the exact version used by the solver, therefore it will only work with a specific version of ghc. I've added two project and freeze files

miguel-negrao:
Sorry, the link was wrong, copy-paste issue, fixed it.

The freeze file will also pin the base version to the exact version used by the solver, therefore it will only work with a specific version of ghc. I've added two project and freeze files for the two ghc versions, and they build correctly. I didn't add a cabal.project.freeze file because then people trying to build the library with `cabal build all` in their machine will not be able to do so unless they have the exact same ghc version, and that would be cumbersome I think. Stack doesn't have this problem as it will also fetch the right ghc as needed.

Does this look good ?

miguel-negrao:
Squashed commits and cache seems to be working. The eolc libraries are always rebuilt but the dependencies are not.

miguel-negrao:
The freeze files were created on my machine with 

```Bash
cabal freeze --project-file=cabal.project.8.8.4 -w ghc-8.8.4
cabal freeze --project-file=cabal.project.8.10.7 -w ghc-8.10.7
```

miguel-negrao:
Removed `stack.yaml.lock` file, not needed anymore.

miguel-negrao:
Currently the CI doesn't build on windows or macos as we would need to conditionally build `essence-of-live-coding-pulse` related packages. I think this would require a different cabal.project file for each OS too. See the log [here](https://github.com/miguel-negrao/essence-of-live-coding/runs/4827282434?check_suite_focus=true).

turion:
Ideally, there would only be one `cabal.project` _text_ file, and the others should be symbolic links to that one. Is that possible?

miguel-negrao:
> Ideally, there would only be one `cabal.project` _text_ file, and the others should be symbolic links to that one. Is that possible?

Done.

turion:
Awesome!

Issue #57, opened 2021-04-22
Add Traversing instance to Cell


turion:
Follow-up to #56.

I think that Cells should be an instance of https://hackage.haskell.org/package/profunctors-5.6.2/docs/Data-Profunctor-Traversing.html#t:Traversing. Basically, this means that there should be a function `traverse' :: Traversable f => Cell m a b -> Cell m (f a) (f b)`. For example, a cell can be lifted into a cell of lists, or a cell of trees.

## To do

* Add `Choice` instance
* Add `Strong` instance
* Add `Traversing` instance, probably implementing `wander` and not `traverse'`

miguel-negrao:
I think `WrappedArrow Cell` is already an instance of Traversing since Cell is an instance of Arrow and ArrowChoice.

turion:
Where is such an instance defined?

miguel-negrao:
Ah, actually, no, not quite. The Choice and Strong we don't have to implement, but we still have to implement the Traversing instance.

miguel-negrao:
I didn't test this, only check that compiled, but I think Cell can be instance of Profunctor, Strong and Choice like below. I'm using coerce just for speed (If I understoddo correctly, coerce is an optimization):

```Haskell
{-# LANGUAGE InstanceSigs #-}
{-# LANGUAGE ScopedTypeVariables #-}

import Data.Profunctor
import Data.Profunctor.Strong
import Data.Profunctor.Choice
import Data.Profunctor.Traversing
import Data.Coerce

instance Monad m => Profunctor (Cell m) where
  dimap :: forall a b c d .(a -> b) -> (c -> d) -> Cell m b c -> Cell m a d
  dimap f g a = coerce $ dimap f g (coerce a :: WrappedArrow (Cell m) b c)

instance Monad m => Strong (Cell m) where
  first' :: forall a b c . Cell m a b -> Cell m (a, c) (b, c)
  first' a = coerce $ first' (coerce a :: WrappedArrow (Cell m) a b) 

instance Monad m => Data.Profunctor.Choice.Choice (Cell m) where
  left' :: forall a b c . Cell m a b -> Cell m (Either a c) (Either b c)
  left' a = coerce $ left' (coerce a :: WrappedArrow (Cell m) a b) 
```
My attempt at a Traversing instance. I'm using the first single state from the original cell to feed the function on the first tick, generating multiple output values. On subsequent ticks we use the multiple states stored in an `f s`.
```Haskell
[...]

Issue #56, opened 2021-04-22
Add Profunctor instance to Cell


turion:
# Short description

A [`Cell`](https://github.com/turion/essence-of-live-coding/blob/e10866d36df22ce0bde4ab98ad64b89e4f99edc3/essence-of-live-coding/src/LiveCoding/Cell.lhs#L87) is a [`Profunctor`](https://hackage.haskell.org/package/profunctors-5.6.2/docs/Data-Profunctor.html#t:Profunctor).

# Explanation

Given functions `y -> a` and `b -> z`, a `Cell m a b` can be transformed into a `Cell m y z`, by precomposing with the first function and postcomposing with the second one.
The step function of a cell has type `cellStep  :: s -> a -> m (b, s)`, so for postcomposition `fmap` is needed.

This shouldn't be to hard, so if you're looking for a way into the code base, feel free to give it a try :)

# To do

* A dependency to the `profunctor` package should be added 
* An `instance Functor m => Profunctor (Cell m)` should be added

PR #19, opened 2020-08-10
Dev investigate speedtest spaceleak

turion:
I discovered that the `SpeedTest` executable has a spaceleak. This merge request removes it, improving performance on the way.

PR #95, opened 2022-01-31
Add  Traversing instance for Cell

miguel-negrao:
Hopefully this solves #56 and #57 

With this instance should `resampleList` and `resampleMaybe` be removed ?

```
> fst <$> steps (traverse' sumC :: Cell IO [Int] [Int]) [[1,2], [1],[3,4]] 
[[0,1],[3],[4,7]]
> fst <$> steps (traverse' sumC :: Cell IO (Maybe Int) (Maybe Int)) [Just 1, Nothing, Nothing, Just 2, Just 3, Nothing] 
[Just 0,Nothing,Nothing,Just 1,Just 3,Nothing]
```


miguel-negrao:
Before the last commit my version of traverse' was not as strict as `resampleList`:

```
> take 4 $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) [[1..],[2],[3]]
[1,2,3,4]
> take 4 $ head $ fst $ runIdentity $ steps (resampleList (arr id) :: Cell Identity [Int] [Int]) [[1..],[2],[3
...
```

After the last commit both of the expressions above result in bottom.

miguel-negrao:
Refactored to its own module `LiveCoding.Profunctors` and added tests.

essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs
```diff
@@ -0,0 +1,81 @@
...
+  {-# INLINE liftA2 #-}
+
+traverseStrict :: (Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
+traverseStrict = (coerce :: ((a -> StrictT f b) -> t a -> StrictT f (t b)) -> (a -> f b) -> t a -> f (t b)) traverse
+{-# INLINE traverseStrict #-}
+
+instance Monad m => Traversing (Cell m) where
+  traverse' (Cell state step) = Cell state step' where
+    step' s a = runStateT (traverse (\a -> StateT (`step` a)) a) s
+  traverse' (ArrM f) = ArrM (traverseStrict f)
```
turion, line 81: The `StrictT` trick is very interesting, but I have a bad feeling to introduce so much generic code which should reside in a separate package, just for this one special case.

I don't understand why `traverseStrict` is needed here, and not in the `Cell` case. And maybe it's possible to implement `traverseStrict` in a simpler way?

Also, I don't understand what exactly is strict here. The container structure we're traversing over? Or the individual values? Because the latter doesn't make sense to me (even though I implemented `resampleList` that way).

turion: > I don't understand why `traverseStrict` is needed here, and not in the `Cell` case. And maybe it's possible to implement `traverseStrict` in a simpler way?

Ah, you said that it's because of strict state! Makes sense.

turion:
I'm unsure about the strictness. I agree that there should be some strictness, but I think the examples shouldn't be so strict. In fact, I'm more and more convinced that I implemented `resampleList` incorrectly.

Some code examples:

```haskell

testSetup x = take 4 $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) x

shouldNotBeBottom = testSetup [[1..],[2],[3]]

shouldBeBottom = testSetup [1 : error "boom", [2], [3]]
...

I think it should only diverge if there is a a problem in the list, but not if there is a problem in the data. That way we can be sure the list is processed completely, but leave it to the cell itself to decide whether it needs to evaluate all the data inside.

Can you add a few tests that capture the expected behaviour? If you use `error` or `undefined` instead of an infinite list, you can use [`assertThrows` or exception handling](https://stackoverflow.com/questions/46330592/is-it-possible-to-assert-an-error-case-in-hunit) to make sure it explodes in the right cases.

File: essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+module LiveCoding.Cell.Profunctor (module X) where
+
+-- base
+import Data.Coerce ( coerce, Coercible )
+import Control.Applicative (liftA2)
+
+-- profunctors
+import Data.Profunctor
+import Data.Profunctor
+    ( Strong, WrappedArrow(WrapArrow), Profunctor )
```
turion, line 16:
```suggestion
    ( Strong, WrappedArrow, Profunctor )
```

essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs
```diff
@@ -0,0 +1,81 @@
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE DeriveAnyClass #-}
+
+-- export only instances
```
turion, line 6: Ok, but we should export them always. I think it's best to avoid the orphan instance and add this instance to `Cell.lhs`.

turion: I mean, you've added it to `LiveCoding.hs`, which is great for users, but it also means that we can't import it internally in the library (because that would be a circular import).

essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs
turion: That line is superseded by the next one I guess?

essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs
```diff
@@ -0,0 +1,81 @@
...
+import Data.Profunctor.Choice ( Choice )
+import Data.Profunctor.Traversing as X
+import Control.Monad.Trans.State (StateT(runStateT, StateT))
+
+-- essence-of-live-coding
+import LiveCoding.Cell ( Cell(Cell, ArrM) )
+
+deriving via (WrappedArrow (Cell m)) instance Monad m => Profunctor (Cell m) 
+deriving via (WrappedArrow (Cell m)) instance Monad m => Strong (Cell m)
+deriving via (WrappedArrow (Cell m)) instance Monad m => Data.Profunctor.Choice.Choice (Cell m)
```
turion, line 27:
```suggestion
deriving via (WrappedArrow (Cell m)) instance Monad m => Choice (Cell m) 
```

turion:
> Hopefully this solves #56 and #57

Yes, it's looking that way!

> With this instance should `resampleList` and `resampleMaybe` be removed ?

No, I think they should be replaced with type-specific aliases `resampleMaybe = traverse'`. The same with `resample`, which works over a `Vector`.

miguel-negrao:
> I'm more and more convinced that I implemented `resampleList` incorrectly.

I had implemented this way to match the strictness in `resampleList`, so that changing to an implementation of `resampleList` based on traverse' would keep the same behaviour.

> I think it should only diverge if there is a a problem in the list, but not if there is a problem in the data. That way we can be sure the list is processed completely, but leave it to the cell itself to decide whether it needs to evaluate all the data inside.

Ok, in that case I will change to using `traverse` and lazy `StateT`.

essence-of-live-coding/src/LiveCoding/Cell/Profunctor.hs
miguel-negrao: So I should move all the code to `Cell.lhs`, correct ?

turion: Yes.

turion:
> Ok, in that case I will change to using `traverse` and lazy `StateT`.

Will that evaluate the list structure strictly? I'd expect it doesn't.

turion:
Some background: I introduced the additional strictness in https://github.com/turion/essence-of-live-coding/pull/19. In fact, we have strictness in the data passing through cells:

```haskell
  Cell state2 step2 . Cell state1 step1 = Cell { .. }
    where
      cellState = Composition state1 state2
      cellStep (Composition state1 state2) a = do
        (!b, state1') <- step1 state1 a
        (!c, state2') <- step2 state2 b
        return (c, Composition state1' state2')
```

This surprises me a lot. I think this should be rethought first.

miguel-negrao:
> Will that evaluate the list structure strictly? I'd expect it doesn't.

The traversal of the traversable structure is lazy when using State.Lazy:

```Haskell
Prelude Control.Monad.Trans.State.Lazy> take 2 $ fst $ flip runState 0 $ traverse (\_ -> modify (+2) >> get) [1..]
[2,4]
```

On the other hand from the code of the composition of Cell above it is probably going to evaluate the whole list before passing it on to the next cell. So the behavior might be different if we just test a single cell or if we test the composition of two cells (i.e. `cell` vs `cell >> (arr id)` ). I will create some tests for this, so we can check on the behavior. 

But I suspect it won't matter much if we use Strict or Lazy State when the cell is composition of two cells...

miguel-negrao:
I've refactored the PR:


- Renamed `Choice` to `CellChoice` so it doesn't clash with profunctors' `Choice`.
- Moved code to `Cell.lhs`
- `traverse'` is using `Control.Monad.Trans.State.Lazy`.
- Reimplemented resampleList, etc as just `traverse'`.
- Added tests for `traverse'`'s lazyness.

This implementation keeps the list created in `resampleList` lazy. This is the case even when composing with another Cell. I suppose that is the case because the bang pattern used in `cell1 >>> cell2` on a list will evaluate the list into either `Nil` or `_ : _ `, which doesn't force the entire list, right ?

Check if this addresses all the issues.

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
+import Control.Exception (ErrorCall)
+import Test.Framework.Providers.HUnit
+import Control.Exception.Base
+import Test.HUnit
 
 test = testGroup "Cell"
   [ testProperty "steps produces outputs"
     $ \(inputs :: [Int]) -> inputs === fst (runIdentity $ steps (id :: Cell Identity Int Int) inputs)
+  , testProperty "traverse' works as expected for Lists"
+  $ \(inputs :: [[Int]]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
```
turion, line 37:
```suggestion
    $ \(inputs :: [[Int]]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
```

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
+  , testProperty "traverse' works as expected for Maybes"
+  $ \(inputs :: [Maybe Int]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testProperty "traverse' for [] is equal to resampleList"
+  $ \(inputs :: [[Int]]) -> fst (runIdentity $ steps (resampleList (arr (+10) :: Cell Identity Int Int)) inputs)
+    === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
```
turion, line 42:
```suggestion
    $ \(inputs :: [[Int]]) -> fst (runIdentity $ steps (resampleList (arr (+10) :: Cell Identity Int Int)) inputs)
      === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs) 
```

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
+  , testProperty "traverse' for Maybe is equal to resampleMaybe"
+  $ \(inputs :: [Maybe Int]) -> fst (runIdentity $ steps (resampleMaybe (arr (+10) :: Cell Identity Int Int)) inputs)
+    === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testCase "traverse' by itself does not force the entire list (ArrM)" $ 
+    testError "Traverse' is too strict" id $ 
+    head $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
```
turion, line 48:
```suggestion
  , testCase "traverse' by itself does not force the entire list (ArrM)" 
    $ testError "Traverse' is too strict" id
    $ head $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
```
And similarly the other cases

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
+import Test.HUnit
 
 test = testGroup "Cell"
   [ testProperty "steps produces outputs"
     $ \(inputs :: [Int]) -> inputs === fst (runIdentity $ steps (id :: Cell Identity Int Int) inputs)
+  , testProperty "traverse' works as expected for Lists"
+  $ \(inputs :: [[Int]]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testProperty "traverse' works as expected for Maybes"
+  $ \(inputs :: [Maybe Int]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testProperty "traverse' for [] is equal to resampleList"
```
turion, line 40: This is given by definition ;)

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
 import qualified Cell.Monad.Trans
+import Control.Exception (ErrorCall)
+import Test.Framework.Providers.HUnit
+import Control.Exception.Base
+import Test.HUnit
 
 test = testGroup "Cell"
   [ testProperty "steps produces outputs"
     $ \(inputs :: [Int]) -> inputs === fst (runIdentity $ steps (id :: Cell Identity Int Int) inputs)
+  , testProperty "traverse' works as expected for Lists"
```
turion, line 36: What's the law you're expecting to hold?

turion:
>     * Renamed `Choice` to `CellChoice` so it doesn't clash with profunctors' `Choice`.

I'd prefer it if `Choice` from `profunctor` were aliased. Does that work as well?

>     * `traverse'` is using `Control.Monad.Trans.State.Lazy`.
>     * Added tests for `traverse'`'s lazyness.
> 
> This implementation keeps the list created in `resampleList` lazy. This is the case even when composing with another Cell. I suppose that is the case because the bang pattern used in `cell1 >>> cell2` on a list will evaluate the list into either `Nil` or `_ : _ `, which doesn't force the entire list, right ?

So an input of `"ok" : error "oh noez"` would not crash? I guess you're right, we don't seq the whole list structure, but only the first constructor when we do the bang pattern on the value passing through the cell.

Can you check whether the `SpeedTest` in `essence-of-live-coding-yampa-speedtest` still doesn't have a space leak? This used to be one of the reasons for the strictness change. (I tried to check with memory profiling, but I couldn't do it.)

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,49 @@ import LiveCoding
...
+    === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testCase "traverse' by itself does not force the entire list (ArrM)" $ 
+    testError "Traverse' is too strict" id $ 
+    head $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
+  , testCase "traverse' by itself does not force the entire list (Cell)" $ 
+    testError "Traverse' is too strict" id $ 
+    head $ head $ fst $ runIdentity $ steps (traverse' cellId :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
+  , testCase "traverse' of a Cell composed with another Cell does not force the entire list" $ 
+    testError "Traverse' is too strict" id $ 
+    head $ head $ fst $ runIdentity $ steps (traverse' cellId >>> cellId  :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
```
turion, line 54: I think you can refactor the tests with `CellSimulation`, they will be easier to understand afterwards :)

essence-of-live-coding/test/Cell.hs
miguel-negrao: Ah, yes, this test was for when resampleList was not just traverse'. Removed these two tests.

essence-of-live-coding/test/Cell.hs
miguel-negrao: But here I'm not checking whether the cell produces a specific value, but just that it doesn't produce an error, so I don't think I can use `CellSimulation` for that, can I ?

miguel-negrao:
> > ```
> > * Renamed `Choice` to `CellChoice` so it doesn't clash with profunctors' `Choice`.
> > ```
> 
> I'd prefer it if `Choice` from `profunctor` were aliased. Does that work as well?

Sure, no problem, will do that :). If the user needs to use profunctor's Choice then they must import qualified by themselves, and profunctor's Choice probably is not going to be used.

> > This implementation keeps the list created in `resampleList` lazy. This is the case even when composing with another Cell. I suppose that is the case because the bang pattern used in `cell1 >>> cell2` on a list will evaluate the list into either `Nil` or `_ : _ `, which doesn't force the entire list, right ?
> 
> So an input of `"ok" : error "oh noez"` would not crash? I guess you're right, we don't seq the whole list structure, but only the first constructor when we do the bang pattern on the value passing through the cell.
> 
Yes, it doesn't crash. The tests that I added are testing this.

```Haskell
  , testCase "traverse' by itself does not force the entire list (Cell)" $ 
    testError "Traverse' is too strict" id $ 
    head $ head $ fst $ runIdentity $ steps (traverse' cellId :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
```

> Can you check whether the `SpeedTest` in `essence-of-live-coding-yampa-speedtest` still doesn't have a space leak? This used to be one of the reasons for the strictness change. (I tried to check with memory profiling, but I couldn't do it.)

I profiled SpeedTest and I didn't see any space leak. The output has a NaN, is that normal ? Looking at the code of SpeedTest, I think that it doesn't use any resample function, so there's no reason why it should change behavior right ?

```
essence-of-live-coding$ stack exec --profile -- SpeedTest +RTS -p -hc
(NaN,NaN,-5.56184561945215e306)
essence-of-live-coding$ stack exec -- hp2ps -e8in -c SpeedTest.hp
essence-of-live-coding$ xdg-open SpeedTest.ps
```

Captura de ecrã de 2022-03-05 18-22-57

essence-of-live-coding/test/Cell.hs
miguel-negrao: I was checking whether traverse' applies the function to each element of the list.

turion:
> > > This implementation keeps the list created in `resampleList` lazy. This is the case even when composing with another Cell. I suppose that is the case because the bang pattern used in `cell1 >>> cell2` on a list will evaluate the list into either `Nil` or `_ : _ `, which doesn't force the entire list, right ?
> > 
> > 
> > So an input of `"ok" : error "oh noez"` would not crash? I guess you're right, we don't seq the whole list structure, but only the first constructor when we do the bang pattern on the value passing through the cell.
> 
> Yes, it doesn't crash. The tests that I added are testing this.
> 
> ```haskell
>   , testCase "traverse' by itself does not force the entire list (Cell)" $ 
>     testError "Traverse' is too strict" id $ 
>     head $ head $ fst $ runIdentity $ steps (traverse' cellId :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
> ```

I think that's not the same thing what I meant. The test I meant would have as input not `[[1, error "Bang !"]]`, but `[1 : error "Bang !"]`. This is subtly different.

> > Can you check whether the `SpeedTest` in `essence-of-live-coding-yampa-speedtest` still doesn't have a space leak? This used to be one of the reasons for the strictness change. (I tried to check with memory profiling, but I couldn't do it.)
> 
> I profiled SpeedTest and I didn't see any space leak.

Awesome, thanks :)

> The output has a NaN, is that normal ?

Yes, I think at some point it exceeds the limits of a `Double`.

> Looking at the code of SpeedTest, I think that it doesn't use any resample function, so there's no reason why it should change behavior right ?

True, I'm puzzled a bit. I thought it used it somewhere, but maybe that was an older temporary version of the test.

Or maybe I simply changed the definition of `>>>`, and thought that I'd have to change the definition of `resampleList` analogously for consistency.

Anyways. Your choice of lazyness seems to be a good one then :) let's keep it like it is now.

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,45 @@ import LiveCoding
...
+  , testCase "traverse' of a Cell composed with another Cell does not force the entire list"
+    $ testError "Traverse' is too strict" id
+    $ head $ head $ fst $ runIdentity $ steps (traverse' cellId >>> cellId  :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
   , Cell.Util.test
   , Cell.Monad.Trans.test
-  ]
+  ]  
+
+testError :: String -> (Bool -> Bool) -> a -> IO ()
+testError message alterBoolValue value = do
```
turion, line 56: `alterBoolValue` is `id` in all usage cases, it seems. Can you remove it?

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,45 @@ import LiveCoding
...
+  let 
+    handler :: ErrorCall -> IO Bool
+    handler e = pure True
+  errored <- catch (pure $ seq value False) handler
+  if alterBoolValue errored then
+    assertFailure message
+  else
+    pure ()
```
turion, line 64: Can you use `when`?

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,45 @@ import LiveCoding
...
+    $ head $ head $ fst $ runIdentity $ steps (traverse' cellId :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
+  , testCase "traverse' of a Cell composed with another Cell does not force the entire list"
+    $ testError "Traverse' is too strict" id
+    $ head $ head $ fst $ runIdentity $ steps (traverse' cellId >>> cellId  :: Cell Identity [Int] [Int]) [[1, error "Bang !"]]
   , Cell.Util.test
   , Cell.Monad.Trans.test
-  ]
+  ]  
+
+testError :: String -> (Bool -> Bool) -> a -> IO ()
```
turion, line 55:
```suggestion
-- tests whether the value `a` throws an exception
testError :: String -> (Bool -> Bool) -> a -> IO ()
```

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,45 @@ import LiveCoding
...
+testError message alterBoolValue value = do
+  let 
+    handler :: ErrorCall -> IO Bool
+    handler e = pure True
+  errored <- catch (pure $ seq value False) handler
```
turion, line 60: I'm confused. Will this be `True` when there was an exception? I believe not.

essence-of-live-coding/test/Cell.hs
turion: So something like `traverse' (cell >>> arr f) = traverse' cell >>> arr (map f)`?

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -117,4 +117,14 @@ test = testGroup "Utility unit tests"
...
+    $ \(inputs :: [[Int]]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+    , testProperty "traverse' works as expected for Maybes"
+    $ \(inputs :: [Maybe Int]) -> fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+    , testProperty "traverse' for [] is equal to resampleList"
+    $ \(inputs :: [[Int]]) -> fst (runIdentity $ steps (resampleList (arr (+10) :: Cell Identity Int Int)) inputs)
+      === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+    , testProperty "traverse' for Maybe is equal to resampleMaybe"
+    $ \(inputs :: [Maybe Int]) -> fst (runIdentity $ steps (resampleMaybe (arr (+10) :: Cell Identity Int Int)) inputs)
+      === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
   ]
```
turion, line 130: But these can be written with `CellSimulation`, right?

essence-of-live-coding/test/Cell.hs
```diff
@@ -23,10 +25,45 @@ import LiveCoding
...
 
 test = testGroup "Cell"
   [ testProperty "steps produces outputs"
     $ \(inputs :: [Int]) -> inputs === fst (runIdentity $ steps (id :: Cell Identity Int Int) inputs)
+  , testProperty "traverse' works as expected for Lists"
+    $ \(inputs :: [[Int]]) -> 
+      fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
+  , testProperty "traverse' works as expected for Maybes"
+    $ \(inputs :: [Maybe Int]) -> 
+      fmap (fmap (+10)) inputs === fst (runIdentity $ steps (traverse' (arr (+10) :: Cell Identity Int Int)) inputs)
```
turion, line 41: Are these tests duplicate to the ones in `test/Cell/Util.hs`?

I think all tests regarding `traverse'` should be in `test/Cell/Util.hs`, and ideally in a separate `testGroup "traverse'"`.

(I know not all tests are sorted like that currently, but they should be ideally.)

miguel-negrao: Ah, yes, has forgotten to delete them in Util.hs, will move tests to Util.hs

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: indeed

essence-of-live-coding/test/Cell.hs
miguel-negrao: Yes, I believe so. If there is no exception in value then `seq value False` returns `False`. If there is an exception then handler will run which will return `True`.

essence-of-live-coding/test/Cell.hs
miguel-negrao: yup

miguel-negrao:
> I think that's not the same thing what I meant. The test I meant would have as input not [[1, error "Bang !"]], but [1 : error "Bang !"]. This is subtly different.

Ah, that is quite subtly different indeed. So one is e 1 : _|_ : Nil and the other 1 : _|_ , right ?

In any case, it also doesn't cause an error. I will change the tests to use 1 : _|_ instead of 1 : _|_ : Nil.

miguel-negrao:
Ok, I squashed the PR as it was too complicated solving conflicts with master otherwise.

Added CellIdentitySimulation to test whether one cell behaves the same as another cell.
Added tests for law `traverse' (arr f) = arr (f <$>)`
Added tests for traverse' with arbitrary Cells created with random state functions (s -> a -> (b,s)) which don't operate on the monad. It's pretty awesome that quickcheck can create random functions !

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -89,9 +103,9 @@ test = testGroup "Utility unit tests"
         $ counterexample labelString
         $ catMaybes inputs === catMaybes outputs
         .||. bufferNotEmpty
-  , testProperty "delay a >>> changes >>> hold a == delay a"
-    $ \(inputs :: [Int]) (startValue :: Int) -> fst (runIdentity $ steps (delay startValue) inputs) ===
-        fst (runIdentity $ steps (delay startValue >>> changes >>> hold startValue) inputs)
+  , testProperty "delay a >>> changes >>> hold a = delay a"
+    $ \(inputs :: [Int]) (startValue :: Int) ->
+      CellIdentitySimulation  (delay startValue) (delay startValue >>> changes >>> hold startValue) inputs
```
turion, line 108:
```suggestion
      CellIdentitySimulation (delay startValue) (delay startValue >>> changes >>> hold startValue) inputs
```

essence-of-live-coding/test/Util.hs
```diff
@@ -81,3 +81,18 @@ instance (Eq b, Show b) => Testable (CellSimulation a b) where
     , output1 = output
     , output2 = []
     }
+
+-- | Basic unit test for 'Cell' identities.
+--   Check whether one cell behaves the same as another cell.
+data CellIdentitySimulation a b = CellIdentitySimulation
+  { cella :: Cell Identity a b
+  , cellb :: Cell Identity a b
+  , commonInput :: [a]
```
turion, line 90:
```suggestion
  { cell1 :: Cell Identity a b
  , cell2 :: Cell Identity a b
  , input :: [a]  
```
So it's more like the other types, and there is no confusion with `a`/`b`

essence-of-live-coding/test/Util.hs
```diff
@@ -81,3 +81,18 @@ instance (Eq b, Show b) => Testable (CellSimulation a b) where
     , output1 = output
     , output2 = []
     }
+
+-- | Basic unit test for 'Cell' identities.
```
turion, line 85: This is an awesome idea! Thanks :)

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
...
+traverseArrLaw f = CellIdentitySimulation (arr (f <$>)) (traverse' (arr f))
+
+traverseCellTest :: forall s a b t . (Traversable t, Data b, Data s) =>
+  s
+  -> (s -> a -> (b,s))
+  -> [t a]
+  -> CellSimulation (t a) (t b)
+traverseCellTest s f input = CellSimulation
+  { cell = traverse' (Cell s (\s a -> pure $ f s a)) :: Cell Identity (t a) (t b)
+  ,  output = runIdentity $ evalStateT (traverse (traverse (\a -> StateT (Identity . (`f` a)) )) input) s
```
turion, line 205:
```suggestion
  , output = runIdentity $ evalStateT (traverse (traverse (\a -> StateT (Identity . (`f` a)) )) input) s
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
...
   ]
+
+testTraverse' :: Test
+testTraverse' = testGroup "Traversing unit tests"
+  [ testProperty "traverse' (arr f) = arr (f <$>) ([a])"
+    $ \(input :: [[Int]], Fn (f :: Int -> Int)) -> traverseArrLaw f input
+  , testProperty "traverse' (arr f) = arr (f <$>) (Maybe a)"
+    $ \(input :: [Maybe Int], Fn (f :: Int -> Int)) -> traverseArrLaw f input
+  , testProperty "traverse' (arr f) = arr (f <$>) (Vecttor 4 Int)"
+    $ \(input :: [(Int,Int,Int,Int)], Fn (f :: Int -> Int)) -> traverseArrLaw f (V.fromTuple <$> input)
```
turion, line 175: You can put these under a common `testGroup`, similarly the other tests.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
...
     }
   ]
+
+testTraverse' :: Test
+testTraverse' = testGroup "Traversing unit tests"
+  [ testProperty "traverse' (arr f) = arr (f <$>) ([a])"
+    $ \(input :: [[Int]], Fn (f :: Int -> Int)) -> traverseArrLaw f input
+  , testProperty "traverse' (arr f) = arr (f <$>) (Maybe a)"
+    $ \(input :: [Maybe Int], Fn (f :: Int -> Int)) -> traverseArrLaw f input
+  , testProperty "traverse' (arr f) = arr (f <$>) (Vecttor 4 Int)"
```
turion, line 174:
```suggestion
  , testProperty "traverse' (arr f) = arr (f <$>) (Vector 4 Int)"
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
     , output = [[0,0,0],[1,1],[2,2,0]]
     }
   ]
+
+testTraverse' :: Test
+testTraverse' = testGroup "Traversing unit tests"
+  [ testProperty "traverse' (arr f) = arr (f <$>) ([a])"
+    $ \(input :: [[Int]], Fn (f :: Int -> Int)) -> traverseArrLaw f input
```
turion, line 171: Instead of replicating the same test for several types, maybe you can also create types

```haskell
data TestInput = IntList [Int] | MaybeInt (Maybe Int) | VectorOfInts (Int, Int, Int, Int)
  deriving (Eq, Show, Arbitrary, CoArbitrary)
```
And then use that type for `input` and `f`.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -8,9 +10,15 @@ import Data.Functor.Identity
...
 import Control.Monad
 import Data.List
+import Control.Exception
 
 -- transformers
 import Control.Monad.Trans.Reader
+import Control.Monad.Trans.State.Lazy
+
+
+-- vector-sized
```
turion, line 20:
```suggestion

-- vector-sized
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
...
+  -> [t a]
+  -> CellSimulation (t a) (t b)
+traverseCellTest s f input = CellSimulation
+  { cell = traverse' (Cell s (\s a -> pure $ f s a)) :: Cell Identity (t a) (t b)
+  ,  output = runIdentity $ evalStateT (traverse (traverse (\a -> StateT (Identity . (`f` a)) )) input) s
+  , ..
+  }
+
+-- | Tests whether the value `a` throws an exception.
+testError :: String  -> a -> IO ()
```
turion, line 210:
```suggestion
testError :: String -> a -> IO ()
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -150,3 +164,58 @@ test = testGroup "Utility unit tests"
...
+  , testProperty "traverse' works as expected for any Cell Identy Int Int created with constructor Cell (Vecttor 4 Int)"
+    $ \(input :: [(Int,Int,Int,Int)], s :: Int, Fn (f :: (Int, Int) -> (Int,Int))) -> traverseCellTest s (curry f) (V.fromTuple <$> input)
+  , testCase "traverse' by itself does not force the entire list (ArrM)"
+    $ testError "traverse' is too strict"
+    $ head $ head $ fst $ runIdentity $ steps (traverse' (arr id) :: Cell Identity [Int] [Int]) [1 : error "Bang !"]
+  , testCase "traverse' by itself does not force the entire list (Cell)"
+    $ testError "Traverse' is too strict"
+    $ head $ head $ fst $ runIdentity $ steps (traverse' cellId :: Cell Identity [Int] [Int]) [1 : error "Bang !"]
+  , testCase "traverse' of a Cell composed with another Cell does not force the entire list"
+    $ testError "Traverse' is too strict"
```
turion, line 189: Ok, so this tests that in fact this throws an exception, or not?

essence-of-live-coding/test/Cell.hs
turion: But that means that the test _fails_ if there is an exception, right? For that use case, no extra function is needed. If an error is raised anywhere in a test, that counts as a test failure. What I wanted was a function where the test _succeeds_ if there is an exception in the value. That way you can test that a function has a certain strictness, by storing an exception somewhere in the place where you expect strictness, and making sure the whole thing explodes. But I think we don't have these places anymore, because `traverse'` is now lazy.

Maybe it's easiest to drop the tests with the errors, since we don't guarantee strictness or laziness anyways.

essence-of-live-coding/test/Util.hs
miguel-negrao: I was actually trying to avoid the names used in the other CellMigrationSimulation, unfortunately this is the dreaded issue of these functions ending up in a common namespace. Is there way around the record issue ?

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: The test succeeds if it does not throw an exception. So in this case as you mentioned elsewhere it is simpler to just test with CellSimulation, since if it is too strict then it will throw an exception and that will consider the test as failed.

essence-of-live-coding/test/Cell.hs
miguel-negrao: > But that means that the test fails if there is an exception, right? 
yes.

> For that use case, no extra function is needed. If an error is raised anywhere in a test, that counts as a test failure.

Ah, I was missing that bit. Then I could change the tests to just a normal CellSimulation. Should I do that, or just drop the tests ?

>  What I wanted was a function where the test succeeds if there is an exception in the value. 

That is just a question of changing the  `errored` to `not errored` in the function `testError` , then it will fail the test if there is no error.

turion: > > For that use case, no extra function is needed. If an error is raised anywhere in a test, that counts as a test failure.
> 
> Ah, I was missing that bit. Then I could change the tests to just a normal CellSimulation. Should I do that, or just drop the tests ?
> 
> > What I wanted was a function where the test succeeds if there is an exception in the value.
> 
> That is just a question of changing the `errored` to `not errored` in the function `testError` , then it will fail the test if there is no error.

I think you can either drop the tests or change them to a `CellSimulation` if you want.

essence-of-live-coding/test/Util.hs
turion: Ah ok I see. I typically don't mind because I rarely use the record functions. I just like record syntax & wildcards, and there it's no trouble if the names are duplicate.

essence-of-live-coding/test/Util.hs
```diff
@@ -81,3 +81,18 @@ instance (Eq b, Show b) => Testable (CellSimulation a b) where
...
+--   Check whether one cell behaves the same as another cell.
+data CellIdentitySimulation a b = CellIdentitySimulation
+  { cella :: Cell Identity a b
+  , cellb :: Cell Identity a b
+  , commonInput :: [a]  
+  }
+
+instance (Eq b, Show b) => Testable (CellIdentitySimulation a b) where
+  property CellIdentitySimulation { .. }
+    = let
```
turion, line 95: Even simpler you can do something like

```haskell
= property CellSimulation { input1 = input, input2 = input, .. }
```

miguel-negrao: But CellSimulation only has cell, input and output . I don't see how the instance can be created from either CellSimulation or CellMigrationSimulation, can it ?

miguel-negrao:
I've added a mechanism to succinctly generate tests for all Traversable instances of interest. It uses induction via type-classes to go through a type-level list a generate a test for each type. If there are any other Traversables that we are interested in, I can add them to the list. Also, I rebased on top of master.

issues:
- I'm getting a warning from haskell language server which I think is bogus:
```
Pattern match(es) are non-exhaustive
In an equation for ‘traverseCellTest’:
    Patterns not matched:
        Proxy _ (Fun _ _) []
        Proxy _ (Fun _ _) (_:_)compile(-Wincomplete-patterns)
```
related to `traverseCellTest _ s (Fn2 f) input = CellSimulation ...` where I'm using the view pattern `Fn`.
- I tried using less Proxies and more type application but I don't know how to tell GHC that a type `a` inside has free type variables, i.e., `a` instead of `Proxy t -> a`. Anyway, it works fine with `Proxy`s.

miguel-negrao:
Since more PRs area appearing, it would be nice to be able to finish this one :-). Our last discussion was about the tests.

miguel-negrao:
> 

Ah, sorry, I think I missed something :-). which changes are you referring to ?

turion:
If you scroll up a bit, you should see a lot of pending discussions :wink: Mostly of stylistic nature.

essence-of-live-coding/test/Cell.hs
miguel-negrao: This function is no longer needed.

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: done

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: done

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: Function was removed.

essence-of-live-coding/test/Util.hs
miguel-negrao: We can't use the same names, so I changed them to cell1', cell2' and input'.

miguel-negrao:
Hi @turion :-). I went through all the discussions above again, there were one or two which were still marked unresolved, and I changed them now to resolved, but in any case I believe my last push from March had hopefully already addressed all the issues (at least that was my intention :-) ). In any case, I might have missed something, off course.

essence-of-live-coding/src/LiveCoding/Cell.lhs
```diff
@@ -412,6 +421,15 @@ Then we can execute the live program in the same way as before.
 
 \begin{comment}
 \begin{code}
+deriving via (WrappedArrow (Cell m)) instance Monad m => Profunctor (Cell m) 
+deriving via (WrappedArrow (Cell m)) instance Monad m => Strong (Cell m)
+deriving via (WrappedArrow (Cell m)) instance Monad m => Data.Profunctor.Choice.Choice (Cell m)
+
+instance Monad m => Traversing (Cell m) where
+  traverse' (Cell state step) = Cell state step' where
+    step' s a = runStateT (traverse (\a -> StateT (`step` a)) a) s
```
turion, line 430:
```suggestion
  traverse' Cell { cellState, cellStep } = Cell
    { cellStep = \s a -> runStateT (traverse (\a -> StateT (`step` a)) a) s
    , ..
    }
```

turion (changes requested):
Ahhh, my bad! I didn't click on "Submit review"! So only i saw my missing comments. Sorry for delaying this for so long.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -168,4 +196,75 @@ test = testGroup "Utility unit tests"
...
+  genTraversableTests f _ = f (Proxy :: Proxy x) : genTraversableTests f (Proxy :: Proxy (TestTraversables xs))
+
+genTraversableTests' :: forall a . GenTests a => 
+  String -> 
+  (forall (t :: * -> *) . (Arbitrary (t Int), Show (t Int), Eq (t Int), Traversable t, Typeable t) => String -> Proxy t -> Test) ->
+  Test
+genTraversableTests' message f = testGroup message $ genTraversableTests (f message) (Proxy :: Proxy a)
+
+instance (Arbitrary a, KnownNat n) => Arbitrary (V.Vector n a) where
+  arbitrary = sequence $ V.replicate arbitrary
```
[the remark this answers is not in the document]
miguel-negrao: Yes, I think mainly the shrink was missing. Will add it.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -168,4 +196,75 @@ test = testGroup "Utility unit tests"
...
+
+cellId :: Monad m => Cell m b b
```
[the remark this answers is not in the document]
turion: Ok :+1:

essence-of-live-coding/src/LiveCoding/Cell.lhs
miguel-negrao: If I apply that change I get the following error:

```
src/LiveCoding/Cell.lhs:431:7: error:
    • Couldn't match expected type ‘Cell m a b’ with actual type ‘s’
      ‘s’ is a rigid type variable bound by
        a pattern with constructor:
          Cell :: forall (m :: * -> *) a b s.
                  Data s =>
                  s -> (s -> a -> m (b, s)) -> Cell m a b,
        in an equation for ‘traverse'’
        at src/LiveCoding/Cell.lhs:429:13-40
    • In the ‘cellState’ field of a record
      In the expression:
        Cell
          {cellStep = \ s a
                        -> runStateT (traverse (\ a -> StateT (`step` a)) a) s,
           ..}
      In an equation for ‘traverse'’:
          traverse' Cell {cellState, cellStep}
            = Cell
                {cellStep = \ s a
                              -> runStateT (traverse (\ a -> StateT (`step` a)) a) s,
                 ..}
    • Relevant bindings include
        cellStep :: s -> a -> m (b, s)
          (bound at src/LiveCoding/Cell.lhs:429:31)
        cellState :: s (bound at src/LiveCoding/Cell.lhs:429:20)
        traverse' :: Cell m a b -> Cell m (f a) (f b)
          (bound at src/LiveCoding/Cell.lhs:429:3)
    |
431 |     , ..
    |       ^^
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -168,4 +196,75 @@ test = testGroup "Utility unit tests"
...
+makeTraversableTest :: forall (t :: * -> *) a . (Testable a, Typeable t) =>  (Proxy t -> a) -> [Char] -> Proxy t -> Test
+makeTraversableTest a message _ = testProperty (message <> " " <> show (typeRep (Proxy :: Proxy t))) (a (Proxy :: Proxy t))
+
+-- | A data type to store types which are instances of 'Traversable'.
+data TestTraversables :: [* -> *] -> *
+
+-- | A type class for induction on the type-level list containing the Traversables. 
+class GenTests a where
+  genTraversableTests ::  
+    (forall (t :: * -> *) . (Arbitrary (t Int), Show (t Int), Eq (t Int), Traversable t, Typeable t) => Proxy t -> Test)
```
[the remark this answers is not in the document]
miguel-negrao: If I change all the occurences to Arbitrary1, Show1, Eq1, etc. And add

```Haskell
instance  (KnownNat n) => Arbitrary1 (V.Vector n) where
  liftArbitrary = V.replicateM
  liftShrink = V.mapM
instance (Arbitrary a, KnownNat n) => Arbitrary (V.Vector n a) where
  arbitrary = arbitrary1
  shrink = shrink1 
```

Then I get this error

```Haskell
test/Cell/Util.hs:231:9: error:
    • Could not deduce (Arbitrary (t Int))
        arising from a use of ‘makeTraversableTest’
      from the context: (Arbitrary1 t, Show1 t, Eq1 t, Traversable t,
                         Typeable t)
        bound by a type expected by the context:
                   forall (t :: * -> *).
                   (Arbitrary1 t, Show1 t, Eq1 t, Traversable t, Typeable t) =>
                   Proxy t -> Test
        at test/Cell/Util.hs:231:9-54
    • In the second argument of ‘($)’, namely
        ‘makeTraversableTest (traverseArrLaw @Int @Int)’
```
I don't entirely undestand how these type classes work (Show1, etc). Do you have an ideia of what need to do more ?

miguel-negrao:
Sorry for leaving this PR abandoned for so long :-). I've tried to address all the points of the previous review. There are still 2 unresolved conversations above, but I've pushed my current version of the code.

essence-of-live-coding/test/Cell/Util.hs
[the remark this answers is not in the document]
miguel-negrao: I had another look at this, and I don't think that `Arbitrary1` will help much here, since there isn't an automatic instance for `Arbitrary (t Int)` even if there is an instance for `Arbitrary1 t`, and `Arbitrary Int`. Such an instance can be made for a specific t easilly (`arbitrary = arbitrary1`), but it is not made automatically it seems, as it would clash with other instances. So I think I would leave it as it is.

essence-of-live-coding/src/LiveCoding/Cell.lhs
miguel-negrao: I don't see a way to get rid of this error, so I think we can leave it as it is ?

miguel-negrao:
Ok, I think I have addressed all the issues of the last review.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -192,4 +212,98 @@ test =
           }
     ]
 
+type TestTraversables' = TestTraversables '[Maybe, [], V.Vector 10, Seq, Map Int]
```
turion, line 215: Can you put everything related to this in a separate module `Cell.Util.Traversable`

turion, line 215: Can we rename:

* `TestTraversables' -> TestTraversables`
* `TestTraversables -> Traversables`

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -192,4 +212,98 @@ test =
...
+          { cell = cellId >>> arr head
+          , input = [1 : error "Bang !"]
+          , output = [1]
+          }
+    ]
+
+traverseArrLaw ::
+  forall a b t.
+  Traversable t =>
+  Proxy t ->
```
turion, line 243: Is it possible to do everything with `TypeApplications` instead of `Proxy`?

essence-of-live-coding/test/Util.hs
```diff
@@ -86,3 +86,20 @@ instance (Eq b, Show b) => Testable (CellSimulation a b) where
...
         }
+
+{- | Basic unit test for 'Cell' identities.
+   Check whether one cell behaves the same as another cell.
+-}
+data CellIdentitySimulation a b = CellIdentitySimulation
+  { cell1' :: Cell Identity a b
+  , cell2' :: Cell Identity a b
+  , input' :: [a]
+  }
```
turion, line 97:
```suggestion
  { cellL :: Cell Identity a b
  , cellR :: Cell Identity a b
  , jointInput :: [a]
  }
```

essence-of-live-coding/test/Cell/Util.hs
turion: If not, no worries.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -192,4 +212,98 @@ test =
...
+        $ makeTraversableTest (traverseCellTest @Int @Int @Int)
+    , testProperty "traverse' by itself does not force the entire list (ArrM)" $
+        CellSimulation
+          { cell = arr head
+          , input = [1 : error "Bang !"]
+          , output = [1]
+          }
+    , testProperty "traverse' by itself does not force the entire list (Cell)" $
+        CellSimulation
+          { cell = cellId >>> arr head
```
turion, line 234:
```suggestion
          { cell = toCell $ arr head
```
And the definition of `cellId` can be deleted.

turion (changes requested):

miguel-negrao:
Thanks for the review ! Will do the changes and report back.

essence-of-live-coding/test/Cell/Util.hs
miguel-negrao: No I don't think it is. I don't think you can do partial application with type application, which I think would be needed...

miguel-negrao:
Hopefully I have addressed the suggestions in the last review. As always, note that when you think this is ready to merge I can first squash que commits, but for now I'm leaving the individual commits to make it easier to see my latest changes.

miguel-negrao:
Hi, perhaps this is ready to merge ?

miguel-negrao:
Great ! It took a while, but is was worth discussing things and polishing the code ! Thanks for your reviews ! :-)

Issue #94, opened 2022-01-26
Support GHC 9.0 and 9.2


turion:

miguel-negrao:
I think we already have support for 9.0 and 9.2, perhaps this can be closed ?

PR #96, opened 2022-01-31
Add support for ghc 9.0.2

miguel-negrao:
Partially solves #94 

Support for 9.2 I think is not possible at the moment because of the [memory package](https://hackage.haskell.org/package/memory). see [here](https://github.com/vincenthz/hs-memory/issues/86).


cabal.project.9.0.2
```diff
@@ -0,0 +1,4 @@
+packages: */*.cabal
+tests: true
+
+with-compiler: ghc-9.0.2
```
turion, line 4: Ah, why is this not needed in the other `cabal.project.*`? There, we just have a symlink to `cabal.project`.

miguel-negrao: I forgot to run the script for the other cabal.project.* files.

The reason to include this is that with the "with-compiler:" flag a user which has multiple ghc versions installed, for instance via ghcup, can just do `cabal build all --project-file=cabal.project.9.02` and cabal will use the right version of ghc, instead of the default ghc. It is not relevant for CI as there there is only one version of ghc available and it is the correct one.

If you agree with this I would then change the the other cabal.project.* files to also pin the ghc version.

turion: Hmmm, I'm undecided on this. Maybe it's a good idea, but it doesn't feel right for me yet. Can we split the PR, and only merge the GHC 9 change, and leave this change here for another time?

turion:
That's great, thanks!

PR #98, opened 2022-02-22
Add Cell.Migrate.NoMigration module

miguel-negrao:
This generalizes the mechanism used by Handles to forgo migration, so that it can be used by other code.

It adds `delay` and `changes` versions that don't require a Data instance.
It adds `arrChangesM` which only runs an action when a value changes and also doesn't require a Data instance. After migration the kleisli function is called again to get a new value.

### DIFF
--- a/essence-of-live-coding/src/LiveCoding/HandlingState.hs
+++ b/essence-of-live-coding/src/LiveCoding/HandlingState.hs
@@ -25,13 +25,10 @@ import LiveCoding.Cell.Monad.Trans
 import LiveCoding.LiveProgram
 import LiveCoding.LiveProgram.Monad.Trans
 
-data Handling h where
-  Handling
-    :: { key    :: Key
-       , handle :: h
-       }
-    -> Handling h
-  Uninitialized :: Handling h
+data Handling h = Handling
+  { key    :: Key
+  , handle :: h
+  }
 
 type Destructors m = IntMap (Destructor m)
 
@@ -153,22 +150,6 @@ destroyUnregistered = do
   put HandlingState { destructors = registered, .. }
 
 -- * 'Data' instances
-
-dataTypeHandling :: DataType
-dataTypeHandling = mkDataType "Handling" [handlingConstr, uninitializedConstr]
-
-handlingConstr :: Constr
-handlingConstr = mkConstr dataTypeHandling "Handling" [] Prefix
-
-uninitializedConstr :: Constr
-uninitializedConstr = mkConstr dataTypeHandling "Uninitialized" [] Prefix
-
-instance (Typeable h) => Data (Handling h) where
-  dataTypeOf _ = dataTypeHandling
-  toConstr Handling { .. } = handlingConstr
-  toConstr Uninitialized = uninitializedConstr
-  gunfold _cons nil constructor = nil Uninitialized
-
 dataTypeDestructor :: DataType
 dataTypeDestructor = mkDataType "Destructor" [ destructorConstr ]
 
--- a/essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
+++ b/essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
@@ -1,0 +1,104 @@
+{-# LANGUAGE Arrows #-}
+{-# LANGUAGE DeriveTraversable #-}
+{-# LANGUAGE RecordWildCards #-}
+
+{-|
+Module      : LiveCoding.Migrate.NoMigration
+Description : Mechanism to save state in a Cell without requiring a Data instance.
+
+If a data type is wrapped in 'NoMigration' then it can be used as the state of a 'Cell'
+without requiring it to have a 'Data' instance. The consequence is that if the type has changed
+in between a livereload, then the previous saved value will be discarded, and no migration attempt
+will happen.
+
+'LiveCoding' does not export 'delay' and 'changes' from this module. These functions should be
+used with a qualified import.
+-}
+module LiveCoding.Migrate.NoMigration where
+
+-- base
+import Data.Data
+    ( Data(dataTypeOf, gunfold, toConstr),
+      Typeable,
+      mkConstr,
+      mkDataType,
+      Constr,
+      DataType,
+      Fixity(Prefix) )
+import Control.Arrow ( Arrow(arr, second), returnA, (>>>) )
+import Control.Monad (guard)
+
+-- essence-of-live-coding
+import LiveCoding.Cell
+import LiveCoding.Cell.Monad ( hoistCellKleisli )
+import qualified LiveCoding.Cell.Feedback as Feedback
+
+-- * 'NoMigration' data type and 'Data' instance.
+
+-- | Isomorphic to @'Maybe' a@ but has a different 'Data' instance. The 'Data' instance for @'NoMigration' a@ doesn't require a 'Data' instance for @a@.
+-- 
+-- If a data type is wrapped in 'NoMigration' then it can be used as the state of a 'Cell'
+-- without requiring it to have a 'Data' instance. The consequence is that if the type has changed
+-- in between a livereload, then the previous saved value will be discarded, and no migration attempt
+-- will happen.
+data NoMigration a = Initialized a | Uninitialized
+    deriving (Show, Eq, Functor, Foldable, Traversable)
+
+fromNoMigration :: a -> NoMigration a -> a
+fromNoMigration _ (Initialized a) = a
+fromNoMigration a Uninitialized = a
+
+dataTypeNoMigration :: DataType
+dataTypeNoMigration = mkDataType "NoMigration" [initializedConstr, uninitializedConstr]
+
+initializedConstr :: Constr
+initializedConstr = mkConstr dataTypeNoMigration "Initialized" [] Prefix
+
+uninitializedConstr :: Constr
+uninitializedConstr = mkConstr dataTypeNoMigration "Uninitialized" [] Prefix
+
+-- |The Data instance for @'NoMigration' a@ doesn't require a 'Data' instance for @a@.
+instance (Typeable a) => Data (NoMigration a)
+  where
+    dataTypeOf _ = dataTypeNoMigration
+    toConstr (Initialized _) = initializedConstr
+    toConstr Uninitialized = uninitializedConstr
+    gunfold _cons nil _ = nil Uninitialized
+
+-- * Utility functions which internally use 'NoMigration'.
+
+-- | Like 'Feedback.delay', but doesn't require 'Data' instance, and only migrates the
+-- last value if it still has the same type.
+delay :: (Monad m, Typeable a) => a -> Cell m a a
+delay a = arr Initialized >>> Feedback.delay Uninitialized >>> arr (fromNoMigration a)
+
+-- | Like 'Utils.changes', but doesn't require Data instance, and only migrates the last
+-- value if it still is of the same type.
+changes :: (Typeable a, Eq a, Monad m) => Cell m a (Maybe a)
+changes = proc a -> do
+  aLast <- delay Nothing -< Just a
+  returnA -< do
+      aLast' <- aLast
+      guard $ a /= aLast'
+      return a
+
+-- | Caching version of 'arrM'. 
+--
+--   Only runs the computation in @m@ when the input value
+--   changes. Meanwhile it keeps outputing the last outputted value. Also runs the computation
+--   on the first tick. Does not require 'Data' instance. On `:livereload` will run action again on
+--   first tick.
+arrChangesM :: (Monad m, Typeable a, Typeable b, Eq a) => (a -> m b) -> Cell m a b
+arrChangesM f = Cell { cellState = Uninitialized, ..}
+  where
+    cellStep Uninitialized        a = h a
+    cellStep (Initialized (a',b)) a = if a == a'
+      then return (b, Initialized (a, b))
+      else h a
+    h a = (\b' -> (b', Initialized (a, b') )) <$> f a
+
+cellNoMigration :: (Typeable s, Functor m) => s -> (s -> a -> m (b, s)) -> Cell m a b
+cellNoMigration state step = Cell { cellState = Uninitialized, ..}
+  where
+    cellStep Uninitialized a = second Initialized <$> step state a
+    cellStep (Initialized s) a = second Initialized <$> step s a
--- a/essence-of-live-coding/test/Cell/Util.hs
+++ b/essence-of-live-coding/test/Cell/Util.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE Arrows #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 module Cell.Util where
 
 -- base
@@ -118,6 +119,22 @@ test = testGroup "Utility unit tests"
     , output1 = [Nothing, Just 2]
     , output2 = [Just 3, Just 4]
     }
+  , testProperty "delay migrates correctly to itself" CellMigrationSimulation
+    { cell1 = LiveCoding.delay 0
+    , cell2 = LiveCoding.delay 0
+    , input1 = [1 :: Int, 2, 3, 4]
+    , input2 = [5 :: Int, 6, 7, 8]
+    , output1 = [0, 1, 2, 3]
+    , output2 = [4, 5, 6, 7]
+    }
+  , testProperty "delay migrates correctly with original type wrapped in data type with single constructor" CellMigrationSimulation
+    { cell1 = LiveCoding.delay 0 :: Cell Identity Int Int
+    , cell2 = arr Stuff >>> LiveCoding.delay (Stuff 99) >>> arr (\(Stuff a) -> a) :: Cell Identity Int Int
+    , input1 = [1, 2, 3, 4] :: [Int]
+    , input2 = [10, 10, 10, 10] :: [Int]
+    , output1 = [0, 1, 2, 3] :: [Int]
+    , output2 = [4, 10, 10, 10] :: [Int]
+    }
   , testProperty "resampleListPar works as expected"
     $ forAll (vector 100) $ \(inputs :: [(Int, Int)]) -> let
         inputs' = fmap pairToList inputs
@@ -150,3 +167,5 @@ test = testGroup "Utility unit tests"
     , output = [[0,0,0],[1,1],[2,2,0]]
     }
   ]
+
+data Stuff a = Stuff a deriving (Eq, Data)
--- a/essence-of-live-coding/test/Main.hs
+++ b/essence-of-live-coding/test/Main.hs
@@ -23,6 +23,7 @@ import qualified Handle
 import qualified Monad
 import qualified Monad.Trans
 import qualified RuntimeIO.Launch
+import qualified Migrate.NoMigration
 
 import LiveCoding
 
@@ -115,6 +116,7 @@ tests =
       ]
     , Cell.test
     , Handle.test
+    , Migrate.NoMigration.test
     , Monad.test
     , Feedback.test
     ]
--- a/essence-of-live-coding/test/Migrate/NoMigration.hs
+++ b/essence-of-live-coding/test/Migrate/NoMigration.hs
@@ -1,0 +1,43 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TupleSections #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+
+module Migrate.NoMigration where
+
+-- base
+import Control.Arrow ( Arrow(arr), (>>>) )
+import Data.Maybe (fromJust)
+import Data.Data (Data)
+
+-- test-framework
+import Test.Framework ( testGroup )
+
+-- test-framework-quickcheck2
+import Test.Framework.Providers.QuickCheck2 ( testProperty )
+
+-- essence-of-live-coding
+import Util
+import qualified LiveCoding.Migrate.NoMigration as NoMigration
+
+data Stuff a = Stuff a deriving (Eq, Data)
+
+test = testGroup "NoMigration unit tests"
+  [ testProperty "LiveCoding.Migrate.NoMigration.delay migrates correctly to itself" CellMigrationSimulation
+    { cell1 = NoMigration.delay 0
+    , cell2 = NoMigration.delay 0
+    , input1 = [1 :: Int, 2, 3, 4]
+    , input2 = [5 :: Int, 6, 7, 8]
+    , output1 = [0, 1, 2, 3]
+    , output2 = [4, 5, 6, 7]
+    }
+  , testProperty "LiveCoding.Migrate.NoMigration.delay different type will not migrate" CellMigrationSimulation
+    { cell1 = NoMigration.delay 0
+    , cell2 = arr Stuff >>> NoMigration.delay (Stuff 99) >>> arr (\(Stuff a) -> a)
+    , input1 = [1 :: Int, 2, 3, 4]
+    , input2 = [10 :: Int, 10, 10, 10]
+    , output1 = [0, 1, 2, 3]
+    , output2 = [99, 10, 10, 10]
+     }
+  ]
--- a/essence-of-live-coding/essence-of-live-coding.cabal
+++ b/essence-of-live-coding/essence-of-live-coding.cabal
@@ -64,6 +64,7 @@ library
     , LiveCoding.Migrate
     , LiveCoding.Migrate.Cell
     , LiveCoding.Migrate.Migration
+    , LiveCoding.Migrate.NoMigration
     , LiveCoding.Migrate.Monad.Trans
     , LiveCoding.Migrate.Debugger
     , LiveCoding.RuntimeIO
@@ -98,6 +99,7 @@ test-suite test
   main-is: Main.hs
   other-modules:
       Cell
+    , Migrate.NoMigration
     , Cell.Monad.Trans
     , Cell.Util
     , Feedback
--- a/essence-of-live-coding/src/LiveCoding.hs
+++ b/essence-of-live-coding/src/LiveCoding.hs
@@ -40,5 +40,6 @@ import LiveCoding.LiveProgram.Monad.Trans as X
 import LiveCoding.Migrate as X
 import LiveCoding.Migrate.Debugger as X
 import LiveCoding.Migrate.Migration as X
+import LiveCoding.Migrate.NoMigration as X hiding (delay, changes)
 import LiveCoding.RuntimeIO as X hiding (update)
 import LiveCoding.RuntimeIO.Launch as X hiding (foreground)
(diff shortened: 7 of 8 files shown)


essence-of-live-coding/src/LiveCoding/HandlingState.hs
```diff
@@ -31,7 +31,6 @@ data Handling h where
        , handle :: h
```
turion, line 31: This can now probably be changed to a regular (non-GADT) definition, and be given a `derive Data`?

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+      mkConstr,
+      mkDataType,
+      Constr,
+      DataType,
+      Fixity(Prefix) )
+
+-- essence-of-live-coding
+import LiveCoding.Cell ( Cell(Cell) )
+import LiveCoding.Cell.Monad ( hoistCellKleisli )
+import Control.Arrow ( Arrow(arr), returnA, (>>>) )
```
turion, line 34: This is an import from `base`, also `Control.Monad` below

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+import qualified LiveCoding.Cell.Feedback as Feedback
+import Control.Monad (guard)
+
+-- |Isomorphic to @'Maybe' a@ but has a different Data instance
```
turion, line 38:
```suggestion
-- | Isomorphic to @'Maybe' a@ but has a different 'Data' instance.
```
Maybe some description what it's used for? Or is the module supposed to be internal?

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+uninitializedConstr :: Constr
+uninitializedConstr = mkConstr dataTypeNoMigration "Uninitialized" [] Prefix
+
+-- |This 'Data' instance is such that deserialization always returns an Uninitialized value.
+instance (Typeable a) => Data (NoMigration a) where
+  dataTypeOf _ = dataTypeNoMigration
+  toConstr (Initialized _) = initializedConstr
+  toConstr Uninitialized = uninitializedConstr
+  gunfold _cons nil _ = nil Uninitialized
+
```
turion, line 61: It would be great to have 2 Haddock section headers, "Data instance", and "Utility functions" (further down)

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+-- Should be identical to traverse'
+resampleNoMigrate :: Monad m => Cell m a b -> Cell m (NoMigration a) (NoMigration b)
+resampleNoMigrate = hoistCellKleisli morph
+  where
+    morph _ s Uninitialized = return (Uninitialized, s)
+    morph singleStep s (Initialized a) = do
+      (!b , s' ) <- singleStep s a
```
turion, line 68: Why is `b` forced? Do you want to make sure `singleStep` is guaranteed to be evaluated? Then maybe `!(b, s')` will do what you want?

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+      return (Initialized b, s')
+
+-- |Doesn't require Data instance, and only migrates the last value if it still has the same type.
```
turion, line 71:
```suggestion
-- | Like 'Feedback.delay', but doesn't require 'Data' instance, and only migrates the last value if it still has the same type.
```

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+delay :: (Monad m, Typeable c) => c -> Cell m c c
+delay a = arr Initialized >>> Feedback.delay Uninitialized >>> arr (fromNoMigration a)
+
+-- |Doesn't require Data instance, and only migrates the last value if it still is of the same type.
```
turion, line 75:
```suggestion
-- | Like 'Utils.changes', but doesn't require Data instance, and only migrates the last value if it still is of the same type.
```

Not sure whether I'm getting haddock links right, though.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+  bLast <- delay Nothing -< Just b
+  returnA -< do
+      bLast' <- bLast
+      guard $ b /= bLast'
+      return b
+
+-- |Efficient version of arrM which only runs the computation in @m@ when the input value
+-- changes. Meanwhile it keeps outputing the last outputted value. Also runs the computation
+-- on the first tick. Does not require Data instance. On livereload will run action again on
+-- first tick.
```
turion, line 87:
```suggestion
-- | Caching version of 'arrM'. 
--
--   Only runs the computation in @m@ when the input value
--   changes. Meanwhile it keeps outputing the last outputted value. Also runs the computation
--   on the first tick. Does not require 'Data' instance. On `:livereload` will run action again on
--   first tick.
```

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+arrChangesM :: (Monad m, Typeable a, Typeable b, Eq a) => (a -> m b) -> Cell m a b
+arrChangesM f = Cell Uninitialized step where
```
turion, line 89:
```suggestion
arrChangesM f = Cell { cellState = Uninitialized, .. }
  where
```

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+    step Uninitialized        a = h a
+    step (Initialized (a',b)) a = if a == a' then return (b, Initialized (a, b)) else h a
```
turion, line 91:
```suggestion
    cellStep (Initialized (a',b)) a = if a == a'
      then return (b, Initialized (a, b))
      else h a
```

essence-of-live-coding/src/LiveCoding/HandlingState.hs
turion: Hah, forget about `derive Data`, the whole point of your MR is that `Handling` doesn't need a `Data` instance anymore :laughing:

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+uninitializedConstr = mkConstr dataTypeNoMigration "Uninitialized" [] Prefix
+
+-- |This 'Data' instance is such that deserialization always returns an Uninitialized value.
+instance (Typeable a) => Data (NoMigration a) where
+  dataTypeOf _ = dataTypeNoMigration
+  toConstr (Initialized _) = initializedConstr
+  toConstr Uninitialized = uninitializedConstr
+  gunfold _cons nil _ = nil Uninitialized
+
+-- Should be identical to traverse'
```
turion, line 62: If it is, why do we need it?

turion: In fact, when do we ever need this function?

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,92 @@
...
+      guard $ b /= bLast'
+      return b
+
+-- |Efficient version of arrM which only runs the computation in @m@ when the input value
+-- changes. Meanwhile it keeps outputing the last outputted value. Also runs the computation
+-- on the first tick. Does not require Data instance. On livereload will run action again on
+-- first tick.
+arrChangesM :: (Monad m, Typeable a, Typeable b, Eq a) => (a -> m b) -> Cell m a b
+arrChangesM f = Cell Uninitialized step where
+    step Uninitialized        a = h a
```
turion, line 90:
```suggestion
    cellStep Uninitialized        a = h a
```

turion (changes requested):
This is a great improvement! Thanks for thinking this through and finding the right spot to position the axe.

Can you add one small unit test for each utility function?

essence-of-live-coding/src/LiveCoding/HandlingState.hs
miguel-negrao: Right, I can change it to non-GADT definition.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
miguel-negrao: Will add description. The module is not internal.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
miguel-negrao: The idea was to take a cell which needs the Data instance and turn it into a Cell which doesn't need the Data instance. I will re-check that it actually does that. 

At the moment traverse' hasn't been merged, so it is not available. The idea would be that once traverse' is merged this can be removed.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
miguel-negrao: This is was basically copy paste from here: https://github.com/turion/essence-of-live-coding/blob/30eb53698242288922ef5601fab69f87b08fc00b/essence-of-live-coding/src/LiveCoding/Cell/Resample.hs#L31-L37 .

miguel-negrao:
> This is a great improvement! Thanks for thinking this through and finding the right spot to position the axe.
> 
Glad to be able to contribute. :-)

> Can you add one small unit test for each utility function?

Yes, will add that, and also incorporate your other feedback where I didn't reply directly.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
turion: Hmm, curious. In that case it would be better to rewrite it like [`resampleMaybe`](https://github.com/turion/essence-of-live-coding/blob/30eb53698242288922ef5601fab69f87b08fc00b/essence-of-live-coding/src/LiveCoding/Cell/Resample.hs#L39).

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
turion: Oh, that's https://github.com/turion/essence-of-live-coding/pull/95 which I've been neglecting. Thanks!

turion: Ok, let's merge #95 first then.

turion: > The idea was to take a cell which needs the Data instance and turn it into a Cell which doesn't need the Data instance. I will re-check that it actually does that.

Ah, yes, if I understand correctly you could check that by implementing some of the utility functions using `traverse'`?

miguel-negrao: I have removed this function. Actually it is not useful. What we want to wrap in `NoMigration` is the state of the Cell and not the input or output value. Unfortunately a Cell with state s can't be converted to a Cell with state `NoMigration s` without requiring a `Data` instance for s, because to build the first Cell the `Data` instance is required.

miguel-negrao:
I have hopefully addressed all your feedback. I have remove the `resampleNoMigration` function, actually it is not useful.

miguel-negrao:
> Can you add one small unit test for each utility function?

Ah, missed this one, will those tests.

miguel-negrao:
While creating the tests for `delay` and `changes` I was quite surprised that changes from `Utils.hs` does not migrate state. Digging a bit more, it seems that if a state is a `Maybe a` and the new state is a `Nothing`, then no migration will ever happen for a previous value of type `Just a`:
 
```Haskell
> data Stuff a = Stuff a deriving (Eq, Data, Show)
> migrate (Nothing) (Just (45::Int))
Nothing
```
This is the case of `changes` which internally has `delay Nothing`.

Does this make sense to you ? If my analysis is correct then I think that the `changes` in Util should not require a `Data` instance for the incoming value, as it is useless since no migration can happen. If this seems ok to you, I would then move the version of changes in this module to `Utils.hs`.

miguel-negrao:
Added tests for `delay`.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
turion: You could still write something like `cellNoMigration :: s -> (s -> a -> m (b, s)) -> Cell m a b`, which has the same type as the `Cell` constructor, but without the `Data` constraint.

essence-of-live-coding/src/LiveCoding/HandlingState.hs
```diff
@@ -25,13 +25,10 @@ import LiveCoding.Cell.Monad.Trans
...
-  Handling
-    :: { key    :: Key
-       , handle :: h
-       }
-    -> Handling h
-  Uninitialized :: Handling h
+data Handling h = Handling
+  {   key    :: Key
+    , handle :: h
+  }
```
turion, line 31:
```suggestion
  { key    :: Key
  , handle :: h
  }
```

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,96 @@
...
+import Control.Monad (guard)
+
+-- essence-of-live-coding
+import LiveCoding.Cell
+import LiveCoding.Cell.Monad ( hoistCellKleisli )
+import qualified LiveCoding.Cell.Feedback as Feedback
+
+-- * 'NoMigration' data type and 'Data' instance.
+
+-- |Isomorphic to @'Maybe' a@ but has a different Data instance. The Data instance for @'NoMigration' a@ doesn't require a 'Data' instance for @a@.
```
turion, line 38:
```suggestion
-- | Isomorphic to @'Maybe' a@ but has a different 'Data' instance. The 'Data' instance for @'NoMigration' a@ doesn't require a 'Data' instance for @a@.
```

essence-of-live-coding/test/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,61 @@
...
+
+-- test-framework-quickcheck2
+import Test.Framework.Providers.QuickCheck2 ( testProperty )
+
+-- QuickCheck
+
+import Util
+import qualified LiveCoding.Migrate.NoMigration as NoMigration
+import LiveCoding ( Data, delay )
+import Data.Maybe (fromJust)
```
turion, line 23: It seems these are not sorted into the libraries

turion (approved):
Great :) can you squash in just some small formatting things?

turion:
> While creating the tests for `delay` and `changes` I was quite surprised that `changes` from `Utils.hs` does not migrate state. Digging a bit more, it seems that if a state is a `Maybe a` and the new state is a `Nothing`, then no migration will ever happen for a previous value of type `Just a`:
> 
> ```haskell
> > data Stuff a = Stuff a deriving (Eq, Data, Show)
> > migrate (Nothing) (Just (45::Int))
> Nothing
> ```
> 
> This is the case of `changes` which internally has `delay Nothing`.
> 
> Does this make sense to you ?

No, this seems like a bug. If the type of `a` doesn't change, the migration should work.

turion:
I tried to reproduce:

```
$ cabal repl
LiveCoding> migrate Nothing $ Just (45 :: Int)
Nothing
```

So I add a test:

```
diff --git essence-of-live-coding/test/Main.hs essence-of-live-coding/test/Main.hs
index 733b83f..4958349 100644
--- essence-of-live-coding/test/Main.hs
+++ essence-of-live-coding/test/Main.hs
@@ -61,6 +61,8 @@ tests =
       $ \x y z -> migrate (Foo2.Fooo z) (Foo1.Foo x y) === Foo2.Foo x
     , testProperty "Finds correct constructor with records"
       $ \barA barB barC baarA baarB -> migrate Foo2.Bar { .. } Foo1.Baar { .. } === Foo2.Baar { .. }
+    , testProperty "Finds correct constructor if type doesn't change"
+      $ \(x :: Int) -> migrate Nothing (Just x) === Just x
     ]
   , testGroup "User migration"
     [ testProperty "Can add migration from Int to Integer"
```

and the test passes!

The reason for this behaviour is the dreaded monomorphization:

```
LiveCoding> migrate (Nothing :: Maybe Int) $ Just (45 :: Int)
Just 45
LiveCoding> migrate (Nothing :: Maybe Bool) $ Just (45 :: Int)
Nothing
```

I don't know what type `ghci` assigns to the `Nothing` in your example, but apparently not `Maybe Int` (why should it). But if you try to write the code out in a file that is compiled, the type is not going to be monomorphized, and there is a type error instead of a silent type assignment.

Adding a test to make sure this was the reason for the `changes` behaviour:

```
diff --git essence-of-live-coding/test/Cell/Util.hs essence-of-live-coding/test/Cell/Util.hs
index dd667d6..60f64d5 100644
--- essence-of-live-coding/test/Cell/Util.hs
+++ essence-of-live-coding/test/Cell/Util.hs
@@ -108,5 +108,13 @@ test = testGroup "Utility unit tests"
         , Just (2 :: Int)
         , Nothing
         ]
    }
+  , testProperty "changes migrates correctly to itself" CellMigrationSimulation
+    { cell1 = changes
+    , cell2 = changes
+    , input1 = [1, 2] :: [Int]
+    , input2 = [3, 4] :: [Int]
+    , output1 = [Nothing, Just 2]
+    , output2 = [Just 3, Just 4]
+    }
   ]
```

And it passes.

turion:
In what situation did you observe the bug? Can you write a test that didn't behave as expected?

miguel-negrao:
> In what situation did you observe the bug? Can you write a test that didn't behave as expected?

So, I've added a test which is failing:

```Haskell
, testProperty "delay migrates correctly with original type wrapped in sum type (Maybe)" CellMigrationSimulation
    { cell1 = LiveCoding.delay 0 :: Cell Identity Int Int
    , cell2 = arr Just >>> LiveCoding.delay (Just 99) >>> arr fromJust :: Cell Identity Int Int
    , input1 = [1, 2, 3, 4] :: [Int]
    , input2 = [10, 10, 10, 10] :: [Int]
    , output1 = [0, 1, 2, 3] :: [Int]
    , output2 = [4, 10, 10, 10] :: [Int]
    }
```

It works with a simple data type like `data Stuff a = Stuff a deriving (Eq, Data)` but not with Maybe... Do you have an idea of what's going on here ?

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
miguel-negrao:
```Haskell
cellNoMigration :: (Typeable s, Functor m) => s -> (s -> a -> m (b, s)) -> Cell m a b
cellNoMigration state step = Cell { cellState = Uninitialized, ..} where
  cellStep Uninitialized a = second Initialized <$> step state a
  cellStep (Initialized s) a = second Initialized <$> step s a

cellNoMigration' :: Monad m => Cell m a b -> Cell m a b
cellNoMigration'(Cell state step) = cellNoMigration state step
cellNoMigration' cell = cell

changes2 :: (Monad m, Eq a, Typeable a) => Cell m a (Maybe a)
changes2 = cellNoMigration' Util.changes
```

```
• Could not deduce (Data a) arising from a use of ‘Util.changes’
  from the context: (Monad m, Eq a, Typeable a)
    bound by the type signature for:
               changes2 :: forall (m :: * -> *) a.
                           (Monad m, Eq a, Typeable a) =>
                           Cell m a (Maybe a)
    at /home/miguel/Development/Haskell/projects/SC/repos/essence-of-live-coding/essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs:111:1-61
  Possible fix:
    add (Data a) to the context of
      the type signature for:
        changes2 :: forall (m :: * -> *) a.
                    (Monad m, Eq a, Typeable a) =>
                    Cell m a (Maybe a)
• In the first argument of ‘cellNoMigration'’, namely
    ‘Util.changes’
  In the expression: cellNoMigration' Util.changes
  In an equation for ‘changes2’:
      changes2 = cellNoMigration' Util.changes
```

If only there was a way to remove the `Data` constraint by some type trickery, or something like `coerce`... I didn't manage to find a way to do it. We want to use the exact same code (`cellStep` function) but without a constraint, which should be safe as we will never require using the associated functions of that constraint.

turion:
> > In what situation did you observe the bug? Can you write a test that didn't behave as expected?
> 
> So, I've added a test which is failing:
> 
> It works with a simple data type like `data Stuff a = Stuff a deriving (Eq, Data)` but not with Maybe... Do you have an idea of what's going on here ?
> 
> I suppose that in general with something like `A a = A1 a | A2 a | A3` a migrate wouldn't know how to migrate from `a` to `A a`, it would have to pick one constructor. In the special case of `A a = A1 a | A2` then there's only one constructor that makes sense to pick.

Yes, so while we can imagine a default migration from `a` to `Maybe a` and other similar types, there is no general mechanism in eolc that detects these cases.

> Perhaps a special migration can be done for these cases, or for a to Maybe a ?

Yes, you probably want to fiddle with https://hackage.haskell.org/package/essence-of-live-coding-0.2.6/docs/LiveCoding-Migrate.html#v:migrateWith and https://hackage.haskell.org/package/essence-of-live-coding-0.2.6/docs/LiveCoding-Migrate-Migration.html#v:userMigration.

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
turion: Yeah, I think this won't work. The `Data` constraint is baked into half of the library, and you'll need it all the time when you call library functions.

If you want to share more code, we can make a new type:

```haskell
data Mealy m s a b = Mealy { s :: s, f :: a -> s -> m (b, s) }
cell :: Data s => Mealy m s a b -> Cell m a b
cell = _
cellNoMigration :: Mealy m s a b -> Cell m a b
cellNoMigration = _
```
Then all library functions that make sense in the "No-Migration-Context" can be written with this.

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -117,4 +118,30 @@ test = testGroup "Utility unit tests"
...
     }
+  , testProperty "delay migrates correctly to itself" CellMigrationSimulation
+    { cell1 = LiveCoding.delay 0
+    , cell2 = LiveCoding.delay 0
+    , input1 = [1 :: Int, 2, 3, 4]
+    , input2 = [5 :: Int, 6, 7, 8]
+    , output1 = [0, 1, 2, 3]
+    , output2 = [4, 5, 6, 7]
+    }
+    , testProperty "delay migrates correctly with original type wrapped in sum type (Maybe)" CellMigrationSimulation
```
turion, line 129:
```suggestion
  , testProperty "delay migrates correctly with original type wrapped in sum type (Maybe)" CellMigrationSimulation
```

essence-of-live-coding/test/Cell/Util.hs
```diff
@@ -117,4 +118,30 @@ test = testGroup "Utility unit tests"
...
+    { cell1 = LiveCoding.delay 0 :: Cell Identity Int Int
+    , cell2 = arr Just >>> LiveCoding.delay (Just 99) >>> arr fromJust :: Cell Identity Int Int
+    , input1 = [1, 2, 3, 4] :: [Int]
+    , input2 = [10, 10, 10, 10] :: [Int]
+    , output1 = [0, 1, 2, 3] :: [Int]
+    , output2 = [4, 10, 10, 10] :: [Int]
+    }
+    , testProperty "delay migrates correctly with original type wrapped in data type with single constructor" CellMigrationSimulation
```
turion, line 137:
```suggestion
  , testProperty "delay migrates correctly with original type wrapped in data type with single constructor" CellMigrationSimulation
```

essence-of-live-coding/test/Main.hs
```diff
@@ -114,6 +115,7 @@ tests =
         }
       ]
     , Cell.test
+    , Migrate.NoMigration.test
     , Handle.test
```
turion, line 119:
```suggestion
    , Handle.test
    , Migrate.NoMigration.test
```

turion (changes requested):
Can you remove the failing test, as everything works as expected?

essence-of-live-coding/test/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    , testProperty "LiveCoding.Migrate.NoMigration.delay different type will not migrate" CellMigrationSimulation
+        { cell1 = NoMigration.delay 0
+        , cell2 = arr Stuff >>> NoMigration.delay (Stuff 99) >>> arr (\(Stuff a) -> a)
+        , input1 = [1 :: Int, 2, 3, 4]
+        , input2 = [10 :: Int, 10, 10, 10]
+        , output1 = [0, 1, 2, 3]
+        , output2 = [99, 10, 10, 10]
+        }
+    ]
```
turion, line 43:
```suggestion
  [ testProperty "LiveCoding.Migrate.NoMigration.delay migrates correctly to itself" CellMigrationSimulation
    { cell1 = NoMigration.delay 0
    , cell2 = NoMigration.delay 0
    , input1 = [1 :: Int, 2, 3, 4]
    , input2 = [5 :: Int, 6, 7, 8]
    , output1 = [0, 1, 2, 3]
    , output2 = [4, 5, 6, 7]
    }
  , testProperty "LiveCoding.Migrate.NoMigration.delay different type will not migrate" CellMigrationSimulation
    { cell1 = NoMigration.delay 0
    , cell2 = arr Stuff >>> NoMigration.delay (Stuff 99) >>> arr (\(Stuff a) -> a)
    , input1 = [1 :: Int, 2, 3, 4]
    , input2 = [10 :: Int, 10, 10, 10]
    , output1 = [0, 1, 2, 3]
    , output2 = [99, 10, 10, 10]
     }
  ]
```

essence-of-live-coding/src/LiveCoding/Migrate/NoMigration.hs
```diff
@@ -0,0 +1,101 @@
...
+
+-- | Caching version of 'arrM'. 
+--
+--   Only runs the computation in @m@ when the input value
+--   changes. Meanwhile it keeps outputing the last outputted value. Also runs the computation
+--   on the first tick. Does not require 'Data' instance. On `:livereload` will run action again on
+--   first tick.
+arrChangesM :: (Monad m, Typeable a, Typeable b, Eq a) => (a -> m b) -> Cell m a b
+arrChangesM f = Cell { cellState = Uninitialized, ..} where
+    cellStep Uninitialized        a = h a
```
turion, line 92:
```suggestion
arrChangesM f = Cell { cellState = Uninitialized, ..}
  where
    cellStep Uninitialized        a = h a
```

miguel-negrao:
Ok, so `where` should always be on the next line. Changed a couple more.

turion:
Thanks a lot!

miguel-negrao:
Thanks a lot for the review ! :)

PR #101, opened 2022-03-24
Investigate the possibility of using ormolu for formatting

miguel-negrao:
This PR is not for merging, just for discussion.

I've applied ormolu v0.4.0.0 to the source code here. @turion have a look at the code and see if you like it.

For the import sections I think we should disable ormolu so that it doesn't reorder imports alphabetically. This can be done with 

```
{- ORMOLU_DISABLE -}

imports here

{- ORMOLU_ENABLE -}
```
Then each group of imports can be manually formatted in vs-code by selecting and doing "format selection". haskell-language-server plugin for vscode supports ormolu.

There's a github action available, so it's very easy to check formatting on push:

```
on: [push, pull_request]
name: build
jobs:
  formatting:
    name: Formatting with ormolu
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2
      - uses: mrkkrp/ormolu-action@v5
  build:
    depends-on: formatting
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        ghc: ['8.8.4','8.10.7']
...
```

Applying ormolu to all files can be done with this script:

```Bash
#!/usr/bin/env bash

check_command() {
    if ! command -v "$1" &> /dev/null
    then
        echo "$1 could not be found"
        exit
    fi
}

check_command git
check_command ormolu
check_command xargs

git ls-files -z '*.hs' | xargs -0 ormolu --mode inplace
```

### DIFF
--- a/demos/Setup.hs
+++ b/demos/Setup.hs
@@ -1,2 +1,3 @@
 import Distribution.Simple
+
 main = defaultMain
--- a/demos/app/DemoSine.hs
+++ b/demos/app/DemoSine.hs
@@ -3,6 +3,7 @@ import LiveCoding
 
 t1 :: Num a => a
 t1 = 4
+
 t2 :: Num a => a
 t2 = 8
 
--- a/demos/app/DemoSineWaitChange.hs
+++ b/demos/app/DemoSineWaitChange.hs
@@ -1,17 +1,18 @@
 -- base
 import Control.Arrow
-
 -- essence-of-live-coding
 import LiveCoding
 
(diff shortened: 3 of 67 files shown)


turion:
I imagine that we could get much further with https://github.com/fourmolu/fourmolu/ because it is configurable.

miguel-negrao:
> I imagine that we could get much further with https://github.com/fourmolu/fourmolu/ because it is configurable.

Indeed. Superseded by #106.

PR #102, opened 2022-03-31
essence-of-live-coding-frp: Continuous time FRP with eolc

miguel-negrao:
This PR contains two packages:

- essence-of-live-coding-frp
- essence-of-live-coding-frp-test

essence-of-live-coding-frp is a port of Bearriver for essence-of-live-coding. 

Notable changes from Bearriver:

- functions which don't use time are of type `Cell m a b`, cells which do are of type `Cell (ClockInfoT m) a b`.
- Event type has been substituted by Maybe.
- Not all Yampa combinators have been ported, notably many switch variants are missing.
- takeWhile and dropWhile implementation was changed, I believe the original ones are buggy. Added tests that check their behaviour.
- Some utility functions which are not present in Bearriver were added. I think of essence-of-live-coding-frp as a package which should have almost everything you need to do FRP programming. Also added a module to automatically uncurry functions and lift them to arrows. Currently the functions added are:
``` 
holdM :: (Monad m, Data a) => Cell m (m a, Maybe a) a`
arrEM ::
  Monad m =>
  (a -> m b) ->
  Cell m (Maybe a) (Maybe b)
catchC :: (Monad m, Data e, Finite e) => Cell (ExceptT e m) a b -> (e -> Cell m a b) -> Cell m a b
throwEC :: Monad m => Cell (ExceptT e m) (Maybe e) ()
dthrowEC :: (Monad m, Data e) => Cell (ExceptT e m) (Maybe e) ()
traverseHold :: (Monad m, Data c) => Cell m a c -> c -> Cell m (Maybe a) c
```
- I changed integral to use the trapezoidal rule instead of the rectangle rule, I think it is more accurate with no drawback. Integral is tested using sine and polynomials.
- I added a function to run ClockInfoT cells which gets the current time using functions from the clock package: `runClockInfoC :: MonadIO m => Cell (ClockInfoT m) a b -> Cell m a b`.
- I did not touch any code in `essence-of-live-coding`, but I think that if this is merged then the time related stuff in `essence-of-live-coding` should be removed. Possibly the function `edge` might not be needed in eolc, and there is a nameclash with the function `once`.

Additionally I added a package essence-of-live-coding-frp-test which implements a Linear Temporal Logic approach to testing FRP, porting the code from the [yampa-test package](https://github.com/ivanperez-keera/Yampa/tree/develop/yampa-test). Some of the tests are ported from that package, but many were created by myself. This is a separate package so that other packages can test their own code using this package. This is due to my assumption that you cannot import library code from a test suite of another package, and it also follows the fact that yampa-test is a separate package from Yampa. 

- This package also has a demo app with a ball bouncing.
- The LTL implementation is different from yampa-tests. Yampa and MSFs are not exactly compatible, they differ in one important aspect. Yampa streams have a value at time 0, which is separate, while MSFs do not. The way `evalT` was implemented in yampa-test and the way some functions are defined in Bearriver break the test suite. I reimplemented evalT directly in terms of functions from eolc. I hope/believe that my re-implementation is correct. I tested with some tautologies from LTL and they hold. Obviously a subtle bug might have been missed.
- I find LTL very useful for testing cells. You can easily test some quite complex conditions.
- tests will fail if passed an empty list. Cells must be evaluated using a non-empty list.
- Technically many cells don't need to be tested with a stream with time values (e.g. all the initialization cells). In order to simplify code and not have additional code, I just test them with time values anyway.

Additional info:
- All code in these packages is formatted using ormolu.
- Tests are passing in CI. 
- Since almost all code is ported from Bearriver and yampa-test, the copyright is attributed to the original authors of the code (Ivan and you). I hope I did that in the correct way.
- This PR needs the `Traversing` instance for `Cell` therefore it goes on top of my PR which introduces that instance.


essence-of-live-coding-frp-test/app/Main.hs
```diff
@@ -0,0 +1,91 @@
...
+    Cell,
+    HandlingStateT,
+    LiveProgram,
+    arrM,
+    foreverE,
+    liftCell,
+    liveCell,
+    liveMain,
+    returnA,
+    (>>>),
```
turion, line 27: Hah, I didn't know this is valid syntax

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
```diff
@@ -0,0 +1,43 @@
...
+
+import Data.Tuple.Curry (Curry (uncurryN))
+import LiveCoding (Arrow (arr), Cell, Typeable, arrChangesM, arrM)
+import LiveCoding.FRP (arrEM)
+
+-- | Version of @arr@ which uncurries functions of arbitrary arity.
+arrUncurry :: (Arrow a, Curry (b -> c) t) => t -> a b c
+arrUncurry = arr . uncurryN
+
+-- | Version of @arrM@ which uncurries functions of arbitrary arity.
```
turion, line 24:
```suggestion
-- | Version of 'arrM' which uncurries functions of arbitrary arity.
```

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
```diff
@@ -0,0 +1,43 @@
...
+arrUncurryM ::
+  (Curry (input -> m output) b, Monad m) =>
+  b ->
+  Cell m input output
+arrUncurryM = arrM . uncurryN
+
+-- | Version of @arrChangesM@ which uncurries functions of arbitrary arity.
```
turion, line 31:
```suggestion
-- | Version of 'arrChangesM' which uncurries functions of arbitrary arity.
```

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
```diff
@@ -0,0 +1,43 @@
...
+-- Functions which are lifted into arrows have to be in completely curried form.
+-- This module provides utility functions to automatically uncurry functions and
+-- lift them to 'Arrow's using the tuple package's 'Curry' typeclass.
+module LiveCoding.FRP.Curry where
+
+import Data.Tuple.Curry (Curry (uncurryN))
+import LiveCoding (Arrow (arr), Cell, Typeable, arrChangesM, arrM)
+import LiveCoding.FRP (arrEM)
+
+-- | Version of @arr@ which uncurries functions of arbitrary arity.
```
turion, line 20:
```suggestion
-- | Version of 'arr' which uncurries functions of arbitrary arity.
```

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
```diff
@@ -0,0 +1,43 @@
...
+-- Description : Function uncurrying utilities
+-- Copyright   : (c) 2022 Miguel Negrão
+-- Maintainer  : miguel.negrao@friendlyvirus.org
+-- Stability   : experimental
+-- Portability : POSIX, Windows
+--
+-- Functions which are lifted into arrows have to be in completely curried form.
+-- This module provides utility functions to automatically uncurry functions and
+-- lift them to 'Arrow's using the tuple package's 'Curry' typeclass.
+module LiveCoding.FRP.Curry where
```
turion, line 14: This should be part of either https://github.com/miguel-negrao/arrow-utils or directly of `essence-of-live-coding`.

essence-of-live-coding/src/LiveCoding/Cell.lhs
```diff
@@ -412,6 +421,15 @@ Then we can execute the live program in the same way as before.
 
 \begin{comment}
 \begin{code}
+deriving via (WrappedArrow (Cell m)) instance Monad m => Profunctor (Cell m) 
+deriving via (WrappedArrow (Cell m)) instance Monad m => Strong (Cell m)
+deriving via (WrappedArrow (Cell m)) instance Monad m => Data.Profunctor.Choice.Choice (Cell m)
+
+instance Monad m => Traversing (Cell m) where
```
turion, line 428: Ah, we have to merge #95 first.

essence-of-live-coding-frp-test/essence-of-live-coding-frp-test.cabal
```diff
@@ -0,0 +1,62 @@
+name:                essence-of-live-coding-frp-test
+version:             0.1.0
+description:         Test framework for essence-of-live-coding-frp using Linear Temporal Logic
+homepage:            https://github.com/turion/essence-of-live-coding
+license:             BSD3
+license-file:        LICENSE
+author:              Ivan Perez, Manuel Bärenz, and Miguel Negrão
+maintainer:          miguel.negrao@friendlyvirus.org
+category:            FRP, Live codingm Bindings
```
turion, line 9:
```suggestion
category:            FRP, Live coding Bindings
```
I think FRP would suffice

essence-of-live-coding-frp-test/essence-of-live-coding-frp-test.cabal
```diff
@@ -0,0 +1,62 @@
+name:                essence-of-live-coding-frp-test
+version:             0.1.0
+description:         Test framework for essence-of-live-coding-frp using Linear Temporal Logic
+homepage:            https://github.com/turion/essence-of-live-coding
+license:             BSD3
+license-file:        LICENSE
+author:              Ivan Perez, Manuel Bärenz, and Miguel Negrão
```
turion, line 7:
```suggestion
author:              Iván Peréz, Manuel Bärenz, and Miguel Negrão
```
And similar places elsewhere.

essence-of-live-coding-frp-test/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2022, Ivan Perez, Manuel Bärenz, and Miguel Negrão
```
turion, line 1: Did you ask Ivan, or are you adding him because of his authorship of `yampa-test`?

.github/workflows/ci.yml
```diff
@@ -20,7 +20,9 @@ jobs:
           enable-stack: true
           stack-version: '2.7'
       - name: Install dependencies
-        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev
+        run: |
+          sudo apt-get update
```
turion, line 24: This is a good idea independently of this PR.

essence-of-live-coding-frp-test/LICENSE
miguel-negrao: I didn't ask, I'm adding due to his authorship of yampa-test and bearriver.

miguel-negrao:
> Thanks a lot! This is a huge contribution, I'll have to come back to it later. If you want to speed it up, maybe we can split it into only the FRP part and the test part?

My pleasure ! :-) Sure, I can do that. So I would do one PR just with the FRP part and another PR (on top of the previous one) with the test stuff ?

Today I was working on my GUI code, and I started wondering if

```Haskell
accum :: (Monad m, Data a) => a -> Cell m (Maybe (a -> a)) (Maybe a)
```

and similar functions should be generalized to 

```Haskell
accum :: (Monad m, Data a, Traversable f) => a -> Cell m (f (a -> a)) (f a)
```

Such that it can either deal with events in the usual sense (`Maybe e`), or with lists of events ( `[e]`). This is relevant for instant when merging events from different sources to be fed into accum. If we want to take just one event then we would use `Maybe e`, but if want to group possible simultaneous events then we can use `[e]`, and the function `accum` can deal with both just fine since it is defined as `traverse' . accumB`.  To me it makes more sense to use the general type signature, although for newcomers it will be harder to understand what it does...

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
miguel-negrao: It has some functions which depend on Cell. Should I put the functions that depend only on Arrow in arrow-utils, and the ones that depend on Cell directly in essence-of-live-coding ?

essence-of-live-coding/src/LiveCoding/Cell.lhs
miguel-negrao: Yes, I use traverse' a lot. Very handy function. :-)

.github/workflows/ci.yml
miguel-negrao: Yes, this is needed independently of this PR. Without it it can't install the linux packages because the server addresses have changed or something like that.

essence-of-live-coding-frp-test/app/Main.hs
miguel-negrao: I generate these lists by accepting hlint suggestions is vscode and then formatted with ormolu. I read somewhere that, if possible, explicit import lists are better because if a package in a new version adds a new function that clashes with another existing function that will not cause a problem.

miguel-negrao:
I've split this into two PRs. This one only has the frp library, the test part will go into another PR. Also I've generalized the functions that use `traverse'` and fixed the documentation issues you mentioned.

essence-of-live-coding-frp/src/LiveCoding/FRP/Curry.hs
turion: Ah I see. Yes, that sounds like the right way to go forward.

essence-of-live-coding-frp/src/LiveCoding/FRP.hs
```diff
@@ -0,0 +1,736 @@
...
+    cellStep (True, _) _ = return (Nothing, (False, 0))
+
+-- | Event source with repeated occurrences with interval q. Note: If the interval is too short w.r.t.
+-- the sampling intervals, the result will be that events occur at every sample. However, no more than
+-- one event results from any sampling interval, thus avoiding an "event backlog" should sampling become
+-- more frequent at some later point in time.
+repeatedly :: (Monad m, Data b) => Time -> b -> Cell (ClockInfoT m) a (Maybe b)
+repeatedly q x
+  | q > 0 = afterEach qxs
+  | otherwise = error "bearriver: repeatedly: Non-positive period."
```
turion, line 314:
```suggestion
  | otherwise = error "essence-of-live-coding-frp: repeatedly: Non-positive period."
```

essence-of-live-coding-frp/src/LiveCoding/FRP.hs
```diff
@@ -0,0 +1,736 @@
...
+joinE :: Maybe a -> Maybe b -> Maybe (a, b)
+joinE = liftA2 (,)
+
+-- | Split event carrying pairs into two events.
+splitE :: Maybe (a, b) -> (Maybe a, Maybe b)
+splitE = NE.unzip
+
+------------------------------------------------------------------------------
+-- Maybe filtering
+------------------------------------------------------------------------------
```
turion, line 505:
```suggestion
-- * Maybe filtering
```

essence-of-live-coding-frp/src/LiveCoding/FRP.hs
```diff
@@ -0,0 +1,736 @@
...
+  where
+    ef = feedback Nothing $ proc (a, me) -> do
+      throwEC -< me
+      liftCell cell -< a
+
+-- Parallel composition and switching
+
+-- Parallel composition and switching over collections with broadcasting
+
+-- #if (  (4) <  4 ||   (4) == 4 && (8) <  14 ||   (4) == 4 && (8) == 14 && (0) <= 3)
```
turion, line 579: Say what now :rofl:

essence-of-live-coding-frp/src/LiveCoding/FRP.hs
```diff
@@ -0,0 +1,736 @@
...
+--   Cell m a (col b)
+-- dpSwitchB sfs sfF sfCs = Cell $ \a -> do
+--   res <- T.mapM (`unMSF` a) sfs
+--   let bs   = fmap fst res
+--       sfs' = fmap snd res
+--   (e,sfF') <- unMSF sfF (a, bs)
+--   ct <- case e of
+--           Maybe c -> snd <$> unMSF (sfCs sfs c) a
+--           Nothing -> return (dpSwitchB sfs' sfF' sfCs)
+--   return (bs, ct)
```
turion, line 600: Do these not work out of the box in eolc?

PR #106, opened 2022-11-16
Dev fourmolu

miguel-negrao:
This PR is not for merging, just for discussion.

I've applied formolu v0.9.0.0 to the source code. I tried to pick all the options of fourmolu which were most similar to the code base here. There are still some differences such as `{ .. }` being turned into `{..}`. @turion have a look at the code and see if you like it. It might be possible to tweak further the config file, but I didn't see anything else obvious.

This supersedes #101.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,13 @@ jobs:
           enable-stack: true
           stack-version: '2.7'
       - name: Install dependencies
-        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev
+        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev wget
+      - name: Install fourmolu
+        run: |
+          wget https://github.com/fourmolu/fourmolu/releases/download/v0.9.0.0/fourmolu-0.9.0.0-linux-x86_64
+          chmod u+x fourmolu-0.9.0.0-linux-x86_64
+      - name: Check formatting with fourmolu
+        run: ./fourmolu-0.9.0.0-linux-x86_64 --mode check $(git ls-files '*.hs')  
       - name: Cabal update
         run: cabal update
       - name: Cache Cabal build artifacts
--- a/essence-of-live-coding-gloss/src/LiveCoding/Gloss/Debugger.hs
+++ b/essence-of-live-coding-gloss/src/LiveCoding/Gloss/Debugger.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE Arrows #-}
+
 module LiveCoding.Gloss.Debugger where
 
 -- base
@@ -6,9 +7,10 @@ import Control.Arrow
 import Data.Data
 
 -- transformers
-import Control.Monad.Trans.Writer.Strict
+
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.State.Strict
+import Control.Monad.Trans.Writer.Strict
 
 -- syb
 import Data.Generics.Text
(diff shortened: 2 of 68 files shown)


miguel-negrao:
Build is failing. Have to check if this is caused by fourmolu...

turion:
Thanks a lot!

> Build is failing. Have to check if this is caused by fourmolu...

Indeed it is. It seems the issue is that fourmolu converts multiline haddocks to single line haddocks, and then chokes on the enumeration. Since the enumeration seems to be correctly rendering haddock, I'd leave it and instead configure fourmolu to use multiline.

fourmolu.yaml
```diff
@@ -0,0 +1,13 @@
...
+import-export-style: diff-friendly
+indent-wheres: true
+record-brace-space: true
+newlines-between-decls: 1
+haddock-style: single-line
+let-style: auto
+in-style: right-align
+respectful: true
+fixities: []
+unicode: never
```
turion, line 13:
```suggestion
function-arrows: trailing
indent-wheres: true
record-brace-space: true
```

We can remove the default values and only leave those that are changed. Also, let's try `function-arrows: trailing`, I think it's more methodical.

turion (changes requested):
Awesome, I like this! In particular, it doesn't mess up the sorted imports, which I find important. Can you rerun with the suggested settings, and if it works we'll merge?

miguel-negrao:
`function-arrows: trailing` is the default, so I left that out. Check if I applied the correct settings. If we manage to get some settings that look ok it would be nice to have this, takes a bit of mental work of our shoulders :-) . If we want we can add a CI step to check that formatting is ok. It is building with cabal locally, let's see how it goes with the CI.

essence-of-live-coding-gloss/src/LiveCoding/Gloss/Debugger.hs
```diff
@@ -1,14 +1,16 @@
...
+
 module LiveCoding.Gloss.Debugger where
 
 -- base
 import Control.Arrow
 import Data.Data
 
 -- transformers
-import Control.Monad.Trans.Writer.Strict
+
```
turion, line 10: This newline is surprising, but not a problem

essence-of-live-coding-pulse-example/app/Main.hs
```diff
@@ -1,23 +1,25 @@
...
+import qualified Data.List.NonEmpty as NonEmpty
 
 import Prelude hiding ((!))
 
 -- transformers
 import Control.Monad.Trans.Reader
 
 -- vector
-import qualified Data.Vector as Vector
+
```
turion, line 20: Again here. Maybe if we remove these newlines it will leave it alone?

turion (approved):
Two small places that could be tweaked, but overall great!

turion:
> If we want we can add a CI step to check that formatting is ok.

Yes, definitely! Ideally, every commit should go through the formatter.

miguel-negrao:
I've added formatting checking to the CI. Not a single misplaced whitespace will get through ever again ! :-D

.github/workflows/ci.yml
```diff
@@ -20,7 +20,13 @@ jobs:
...
           stack-version: '2.7'
       - name: Install dependencies
-        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev
+        run: sudo apt-get install -y libxml2-utils libgl1-mesa-dev libglu1-mesa-dev freeglut3-dev libpulse-dev libblas-dev liblapack-dev libasound2-dev wget
+      - name: Install fourmolu
+        run: |
+          wget https://github.com/fourmolu/fourmolu/releases/download/v0.9.0.0/fourmolu-0.9.0.0-linux-x86_64
+          chmod u+x fourmolu-0.9.0.0-linux-x86_64
+      - name: Check formatting with fourmolu
+        run: ./fourmolu-0.9.0.0-linux-x86_64 --mode check $(git ls-files '*.hs')
```
turion, line 29: Maybe it's possible to install fourmolu via cabal?

turion, line 29: We have `*.lhs` files as well :see_no_evil:

turion: But let's not have fourmolu format those, they are specifically formatted to fit well into the article.

turion: Argh. It seems if we do that, it installs an older fourmolu to be compatible with GHC 8.8.4. That does other formatting. It would be ok to drop 8.8 if GHC 9.2. and 9.4 were working already, but I'm stuck with #108 .

turion:
Thanks a lot!

